PR #3, opened 2021-05-25
Feature/wallets api endpoints

morrieinmaas:
MVP for wallets endpoint


app/routers/wallet.py
```diff
@@ -0,0 +1,314 @@
...
+    except Exception as e:
+        await aries_agent_controller.terminate()
+        logger.error(f"The following error occured:\n{e!r}")
+        raise HTTPException(
+            status_code=500,
+            detail=f"Something went wrong: {e!r}",
+        )
+
+
+@router.get("/", tags=["wallets"])
```
vineeth14, line 141: Could name it /wallet/create_wallet

morrieinmaas: hmm not sure, cos this is the default rout under wallets like `/` for a domain. So I'd say this should either go (aka be absent) or some helpful content. I would'nt call this one create_wallet because nothing is created 

or do you, @vineeth14,  mean the post method in line 155? then that'd make sense! Pushed changes for that 🙌

vineeth14: Yeah meant the post method, my bad with the comment placement

vineeth14 (commented):
I think this should mention which attributes are required. @morrieinmaas
Maybe mention that the wallet name should be unique as well.

app/routers/wallet.py
```diff
@@ -0,0 +1,314 @@
...
+
+# TODO: This should be somehow retsricted?!
+@router.post("/create-wallet", tags=["wallets"])
+async def create_wallet(wallet_payload: dict = None):
+    """
+    Create a new wallet
+
+    Parameters:
+    -----------
+
```
vineeth14, line 162: I think this should mention which attributes are required. @morrieinmaas
Maybe mention that the wallet name should be unique as well.

morrieinmaas: hmm, well for now it's a dict really - there is an example in the docstring. I reckon that's a good place to use a schema (model) instead

PR #7, opened 2021-06-01
Feature/auth

morrieinmaas:
Add some conditional building blocks for instantiating the AriesAgent or multitenant controller based on passing headers


morrieinmaas:
When this is merged another task is to implement that auth for all endpoints retrospectively (aka the ones we already have) and going forward for all the ones to come. 

Also, if you have a better idea/way of implementing this please do mention.

morrieinmaas:
maybe one thing to consider is whether we want to put all the private functions/business logic into utils so they are a collection of things in one place and they can be used/pulled in in other routers/places also.

app/routers/wallet.py
```diff
@@ -16,15 +18,106 @@
...
 
 
-@router.get(
-    "/create-pub-did", tags=["did"], response_model=DidCreationResponse
-)
-async def create_public_did():
+async def __create_did(controller):
+    generate_did_res = await controller.wallet.create_did()
+    if not generate_did_res["result"]:
+        controller.terminate()
```
michaelwiles, line 31: It is not an issue for this actual code - actually for the cloudcontroller context... We should see if we can use the context manager pattern in python to manage the controller... 

Then thix contesxt would go
```
with controller.wallet.create_did() as generate_did_res: 
    // do the work with created did
```

That way the controller wallet can manage it's own termination etc

If we did that this code would not need to know it needs to terminate. IT would just use the with context and that would be handled

michaelwiles (commented):
So I'd like to, if possible, abstract out all the requests calls... i.e. all the http calls or the remote calls should be done by some kind of delegate. So create something like [WalletController](https://github.com/didx-xyz/aries-cloudcontroller-python/blob/main/aries_cloudcontroller/controllers/wallet.py)

So we can then the normal code does not depend on what mechanism we're using to call out...

There's a bunch of strengths: 
1. we get to change the library that we use for doing the remote calls without affecting any of the "business logic"
2. we have a good idea of what we're calling when - so when we need to make the same call again we don't rewrite the code
3. If we need to add error handling or other things, maybe logging or auditing we can
4. Mocking also becomes a lot easier because now we have something to mock.

So I"m suggesting we actually call these xxx_delegate and arguably make them classes. But _maybe_ we stick with the existing pattern in cloudconroller whereby they are called controllers and use the existing code constructs.

app/routers/wallet.py
```diff
@@ -65,77 +147,30 @@ async def create_public_did():
...
-                detail=f"Something went wrong. Could not obtain public DID. {error_json}",
-            )
+        await __post_to_ledger(url, payload)
+
+        TAA = await __get_taa(controller)
+
+        await __accept_taa(controller, TAA)
+
+        await __assign_pub_did(controller, did_object)
+
```
michaelwiles, line 157: So here we have multiple async calls but they're effectively run synchronously as each one waits for the next one to finish before continuing - but between them there isn't too much shared state...

So could we not call __accept_taa and __assign_pub_did  in tasks and then wait for the tasks to finish... so we won't have to wait for _accpt_taa to finish before running __assign_pub_did?

Having said that there might be some dependency on the back end between them which would make this impossible. I'm just raising the possibility.

So this would become:

```
accept_taa = asyncio.create_task(__accept_taa(controller, TAA))
assign_pub_did = asyncio.create_task(__assign_pub_did(controller, did_object))
await accept_taa
await assign_pub_did
```

So the create_task call will start the method running asynchronously causing each of the __accept_taa and __asign_pub_did to be run asynchrously.

Then awaiting both will ensure that after the second await but have finished before continuing.

michaelwiles: But of course, this is only possible if these tasks can be run concurrently - might be some reason why they can't.

app/routers/wallet.py
morrieinmaas: yeah , context manager could be good here, but If I'm not missing the obvious that's require changing the cloudcontroller to have `__enter__` and `__exit__` methods?

Also:
`with controller.wallet.create_did() as generate_did_res:` needs ot at least be async with

I guess we could also create a helper class that just instatiates the controller in a contextmanager capapble fashion?

morrieinmaas:
> So I'd like to, if possible, abstract out all the requests calls... i.e. all the http calls or the remote calls should be done by some kind of delegate. So create something like [WalletController](https://github.com/didx-xyz/aries-cloudcontroller-python/blob/main/aries_cloudcontroller/controllers/wallet.py)
> 
> So we can then the normal code does not depend on what mechanism we're using to call out...
> 
> There's a bunch of strengths:
> 
>     1. we get to change the library that we use for doing the remote calls without affecting any of the "business logic"
> 
>     2. we have a good idea of what we're calling when - so when we need to make the same call again we don't rewrite the code
> 
>     3. If we need to add error handling or other things, maybe logging or auditing we can
> 
>     4. Mocking also becomes a lot easier because now we have something to mock.
> 
> 
> So I"m suggesting we actually call these xxx_delegate and arguably make them classes. But _maybe_ we stick with the existing pattern in cloudconroller whereby they are called controllers and use the existing code constructs.

Sorry I'm not sure I can follow here. Are you suggesting to make a controller for the controller?

app/utils.py
```diff
@@ -0,0 +1,33 @@
...
+
+import os
+from typing import Optional
+
+admin_url = os.getenv("ACAPY_ADMIN_URL")
+admin_port = os.getenv("ACAPY_ADMIN_PORT")
+is_multitenant = os.getenv("IS_MULTITENANT", False)
+
+
+def create_controller(req_header: Header):
```
michaelwiles, line 12: nice - we want to use this pattern more and maybe even call it a factory (potentially)...

app/routers/wallet.py
michaelwiles: yes - would only do it if it can be done inside the cloudcontroller library. Would strongly prefer not to wrap it in this context. Probably best if we introduce another method that creates it as a context manager

app/routers/wallet.py
```diff
@@ -16,15 +18,106 @@
...
+        raise HTTPException(
+            # TODO: Should this return HTTPException, if so which status code?
+            # Check same for occurences below
+            status_code=404,
+            detail=f"Something went wrong.\nCould not generate DID.\n{generate_did_res}",
+        )
+    return generate_did_res
+
+
+async def __post_to_ledger(url, payload):
```
michaelwiles, line 41: We create a ledger service or ledger delegate (not convinced it should be called controller as that's now what it is) and put this method there...

So the wallet context then doesn't know how to call the ledger, it just calls a method post_to_ledger on the ledger service/delegate.

morrieinmaas:
cool thanks for the feedback I'll go through it asap

michaelwiles:
> > So I'd like to, if possible, abstract out all the requests calls... i.e. all the http calls or the remote calls should be done by some kind of delegate. So create something like [WalletController](https://github.com/didx-xyz/aries-cloudcontroller-python/blob/main/aries_cloudcontroller/controllers/wallet.py)
> > So we can then the normal code does not depend on what mechanism we're using to call out...
> > There's a bunch of strengths:
> > ```
> > 1. we get to change the library that we use for doing the remote calls without affecting any of the "business logic"
> > 
> > 2. we have a good idea of what we're calling when - so when we need to make the same call again we don't rewrite the code
> > 
> > 3. If we need to add error handling or other things, maybe logging or auditing we can
> > 
> > 4. Mocking also becomes a lot easier because now we have something to mock.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So I"m suggesting we actually call these xxx_delegate and arguably make them classes. But _maybe_ we stick with the existing pattern in cloudconroller whereby they are called controllers and use the existing code constructs.
> 
> Sorry I'm not sure I can follow here. Are you suggesting to make a controller for the controller?

So...
```
async def __post_to_ledger(url, payload):
    r = requests.post(url, data=json.dumps(payload), headers={})
    if r.status_code != 200:
        error_json = r.json()
        raise HTTPException(
            status_code=r.status_code,
            detail=f"Something went wrong.\nCould not write to Ledger.\n{error_json}",
        )
    return r

```
This method, since all it's doing is making a remote call, will move to ledger_service or ledger_delegate (not convinced it should be called controller as that's not what it is - it's not controlling anything, it's just delegating) .

And then it will be called by the wallet context needs to post to the ledger.

app/routers/wallet.py
morrieinmaas: Ah I think I see what you mean. In general, oncurrent will be useful when multiple endpoint calls are made from a multitude of clients so they don't have to wait. But as you correctly assume in this case (kinda per endpoint if you so will) this cannot be concurrent in most cases, because eg here the TAA must have been accepted and the ledger must be aware of that for the controller to be able to write a did. So if we made that cuncurrent we'd get an error if we try writing a did before accepting TAA has finished successfully. But, again for endpoints as a whole, I surely agree, that concurrency will be good

app/utils.py
morrieinmaas: Yeah something like this. I reckon even making this a context managable class might work here. (the reason being that we otherwise have to decide to make a breaking change to cloudcontroller and I'm not sure Will wants this in the controller)

michaelwiles:
> > So I'd like to, if possible, abstract out all the requests calls... i.e. all the http calls or the remote calls should be done by some kind of delegate. So create something like [WalletController](https://github.com/didx-xyz/aries-cloudcontroller-python/blob/main/aries_cloudcontroller/controllers/wallet.py)
> > So we can then the normal code does not depend on what mechanism we're using to call out...
> > There's a bunch of strengths:
> > ```
> > 1. we get to change the library that we use for doing the remote calls without affecting any of the "business logic"
> > 
> > 2. we have a good idea of what we're calling when - so when we need to make the same call again we don't rewrite the code
> > 
> > 3. If we need to add error handling or other things, maybe logging or auditing we can
> > 
> > 4. Mocking also becomes a lot easier because now we have something to mock.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So I"m suggesting we actually call these xxx_delegate and arguably make them classes. But _maybe_ we stick with the existing pattern in cloudconroller whereby they are called controllers and use the existing code constructs.
> 
> Sorry I'm not sure I can follow here. Are you suggesting to make a controller for the controller?

see here https://github.com/didx-xyz/aries-cloudapi-python/pull/7/files/227d7ba97d090d16536a94326e2d5a8c1c821841#r644683337

app/routers/wallet.py
wip-abramson: Hey looks cool, just having a quick scan. 

assign_pub_did has a dependency on accept_taa. It will fail if the DID on the ledger has not accepted the TAA.

morrieinmaas:
@wip-abramson also please look at the latest version of changes

app/utils.py
michaelwiles: So the ClientSession (which is what is closed by the terminate) already supports context managers etc...

and furthermore, adding the methods would not break any clients already using the default approach - they wouldn't call the methods.

Reality is it would improve the quality and usability of these Controllers as the clients would no longer need to know to call terminate when they're done.

morrieinmaas:
sorry, not sure what you're trying to get at here https://github.com/didx-xyz/aries-cloudapi-python/pull/7#discussion_r645639594 (or whether that is outdated or still relevant?

app/routers/wallet.py
michaelwiles: hmmm made a mistake with my code example...

would be something more like 

```
with get_wallet_controller() as wallet_controller:
 generate_did_res = await controller.create_did()
 // carry on with processing

```

app/routers/wallet.py
```diff
@@ -16,15 +18,106 @@
...
 ledger_url = os.getenv("LEDGER_NETWORK_URL")
 
 
-@router.get(
-    "/create-pub-did", tags=["did"], response_model=DidCreationResponse
-)
-async def create_public_did():
+async def __create_did(controller):
+    generate_did_res = await controller.wallet.create_did()
+    if not generate_did_res["result"]:
```
michaelwiles: So I'd really like to see if we can move this validation code somewhere else... 
Either we:
1. go through some analysis and decide how important it actually is - on a case by case basis and thus we can potentially remove it in some cases.
2. remove it altogether but make sure we throw the best error we can at the end (i.e. we make sure that our api responds with something potentially relevant). Error handling is _hard_ I know.
3. embed it in the wallet controller. This would be first prize though not sure the cloudcontroller people would be keen. They haven't done it so far so not sure they'll do it now
4. Create our own wrapper or something else round the controllers to centralise and manage this validation code. 

Number 4 is arguably the most viable and it's not a great deal of work. Just a lot of typing.

morrieinmaas:
This should be good to merge now

app/facade.py
```diff
@@ -0,0 +1,259 @@
...
+        yield controller
+    except Exception as e:
+        # We can only log this here and not raise an HTTPExeption as
+        # we are past the yield. See here: https://fastapi.tiangolo.com/tutorial/dependencies/dependencies-with-yield/#dependencies-with-yield-and-httpexception
+        logger.error(f"{e!r}")
+    finally:
+        await controller.terminate()
+
+
+async def create_did(controller):
```
michaelwiles, line 66: I'd be keen to start adding type annotations to these kinds of methods... will help the dev who uses it to know what they're dealing with (And in fact ide's can utilise that to help with code insight) - not necessary for this review cycle, just flagging it for future

app/facade.py
```diff
@@ -0,0 +1,259 @@
...
+    payload: dict
+        The payload to be posted of the form:
+        {
+            "network": "stagingnet",
+            "did": did_object["did"],
+            "verkey": did_object["verkey"],
+            "paymentaddr": "somestring",
+        }
+
+    Returns:
```
michaelwiles, line 107: would be keen to introduce a _type_ to hold the values that are returned. Especially now that we have this facade we can surface to the dev who uses this what they can expect - don't think it'd be very much work to be honest

PR #10, opened 2021-06-06
Fix line endings.

WadeBarnes:
Signed-off-by: Wade Barnes <wade@neoterictech.ca>

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,13 @@
+# Set the default behavior, in case people don't have core.autocrlf set.
+* text=auto
+
+# Declare files that will always have LF line endings on checkout.
+*.sh text eol=lf
+*.md text eol=lf
+*.json text eol=lf
+*.yaml text eol=lf
+*.yml text eol=lf
+*.conf text eol=lf
+.env* text eol=lf
+*.env text eol=lf
+manage text eol=lf


michaelwiles (commented):
Is there a good reason for customising the line endings?

Can't we just make all text endings = lf ?

WadeBarnes:
Yes, by default on Windows line endings will be converted automatically to `CrLf` (the default behavior of Git is `* text=auto`).  This in turn causes issues for the Docker builds.  Files get added to the images with the `CrLf` line endings, which then causes issues when running the containers.  This surfaces most readily with startup scripts.  The `CrLf` line endings cause 'access denied' errors when they are executed.

WadeBarnes:
If you're suggesting setting `* text eol=lf` and being done with it, that is acceptable too.

morrieinmaas:
@WadeBarnes sorry this took so long. Merged from [here](https://github.com/didx-xyz/aries-cloudapi-python/pull/24) because sonar cloud isn't happy being triggered from fork.

michaelwiles:
In the interests of full disclosure @WadeBarnes the issue stems from the fact that we now run sonar on our builds and that requires a sonar token to be configured on the repository (in repository settings). It is not configured in your repository so the build failed. 

I'll do some follow up in this regard to see what can be done to address this.

michaelwiles:
see [here](https://jira.sonarsource.com/browse/MMF-1371) for the feature request to address this...

Issue #9, opened 2021-06-04
Build reference architecture for create_public_did call on wallet


michaelwiles:
I'm going to take this call and flesh out a suggested reference architecture for us to shake a stick at (critique)

morrieinmaas:
@michaelwiles @vineeth14 Not really sure what is left on this issue? Please clarify and I'll do the changes 🙏

mrcbond88:
We think this might be done already. @michaelwiles please can you confirm for @morrieinmaas

mrcbond88:
@michaelwiles We need to get this bad boy merged in please? Seems it's causing a bit of a roadblock. Once it's in we'll also need a fair amount of refactoring.

morrieinmaas:
@michaelwiles I had to unmerge this, because the proposed changes break for instance the `/wallets/create-pub-did` endpoint, for a start. I haven't tried more endpoints but I suspect there will be other errors. 

Also, should've maybe commented on that earlier but overlooked that somehow. I don't think most people using this right now want daemon mode for the compose startup. Hence, I'd not make the changes to the start/up section in manage script. At least at this stage the terminal output comes in very handy. If you are unhappy with this spamming your terminal you could just open another shell or defer it to the background eg with `nohup ./manage.sh start &`.

michaelwiles:
what is the issue with the /wallets/create-pup-did endpoint? are you talking about the change to the api key? I thought you asked that it is received from the client? 

I'm happy to put that back

morrieinmaas:
Closed by #11

PR #11, opened 2021-06-07
#9 Refactorings for reference architecture

michaelwiles:
I took the wallet.create_public_did and refactored it to create a pattern for us going forward.

To summarise:
1. introduced a "core" context which is the "core" processing of our application (no decorators or fastapi "annotations"). Not a great name but not sure what else to call this.
2. removed the router into its own context - it provides the entry point into core
3. created delegates for the remote (http) calls. Don't think it necessary to add this for calls to the agent controller
4. created an agent factory for creating the agent controller

What I'm trying to do is create smaller pieces of functionality and knowledge - that is that classes/functions know only what they need to know and don't know "too much". Wanting to make the code more "[SOLID](https://en.wikipedia.org/wiki/SOLID#:~:text=In%20object%2Doriented%20computer%20programming,understandable%2C%20flexible%2C%20and%20maintainable.&text=The%20Interface%20segregation%20principle%3A%20%22Many,one%20general%2Dpurpose%20interface.%22)".

I welcome any thoughts/comments/suggestions or push back... let's pool our skills and knowledge to create something good.

I still want to add some tests to this, but that will probably be in a different pull request.

### DIFF
--- a/.github/workflows/continuous-build.yml
+++ b/.github/workflows/continuous-build.yml
@@ -28,6 +28,9 @@ jobs:
         python -m pip install --upgrade pip
         pip install pytest
         if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+    - name: Start Test Harness
+      run: |
+        manage up
     - name: Test with pytest
       run: |
         pytest --cov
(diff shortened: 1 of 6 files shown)


michaelwiles:
this pull request became superfluous as @morrieinmaas did the refactoring in feature/auth PR

PR #19, opened 2021-06-11
Feature/9 refactor get public did call (renamed from reference architecture

michaelwiles:
Included here in is rename and group the facades - introduced acapy_ledger_facade and acapy_wallet_facade

We need to do some maturation of the mechanism for locating the aries agent controller.

So I apologise profusely as along with this issue was _also_ the issue of the github actions maturation. This has meant he inclusion of a _lot_ of extra commits (to get the github builds to work) which I should have removed but I'd already merged from development when I realised this (I needed to rebase these commits before merge from development).

So focus on the commits _after_ the merge from development as these ones refactor the get public did call to be inline with the reference architecture done by @morrieinmaas


.github/workflows/continuous-build.yml
```diff
@@ -21,20 +21,31 @@ jobs:
...
+        ./manage up
+       echo waiting for dockers to startup, sleeping for 60 seconds
+       sleep 60
+       echo wait over, continuing
+    - name: Install test dependencies.
+      run: |
+        cd app
         python -m pip install --upgrade pip
-        pip install pytest
+        pip install pytest pytest-custom_exit_code
```
michaelwiles, line 35: this is relevant as other wise pytest exits with 1 when tests fail so the build does not continue - we want the build to continue as we want sonar to run even when tests fail.

app/acapy_wallet_facade.py
```diff
@@ -0,0 +1,84 @@
+import logging
```
michaelwiles, line 1: see comment for acapy_ledger_facade

app/core/wallet.py
```diff
@@ -0,0 +1,59 @@
+from typing import Optional
```
michaelwiles, line 1: introduced a core wallet call

michaelwiles: @morrieinmaas thanks for your changes - look how much more lightweight this call is!

app/facade.py
```diff
@@ -3,20 +3,22 @@
...
 
+admin_api_key = os.getenv("ACAPY_ADMIN_API_KEY")
 
 logger = logging.getLogger(__name__)
 
 
+# TODO: address the way the controller is created
 @asynccontextmanager
-async def create_controller(req_header: Header):
+async def create_controller(place_holder) -> AriesAgentControllerBase:
```
michaelwiles, line 21: need to do some work in this context - defoes need to introduce that fast api dependency

app/facade.py
```diff
@@ -59,35 +63,14 @@ async def create_controller(req_header: Header):
...
         # we are past the yield. See here: https://fastapi.tiangolo.com/tutorial/dependencies/dependencies-with-yield/#dependencies-with-yield-and-httpexception
+        # yes, but we are currently _not_ being called by a fast api dependency and besides. And I want to throw the
+        # exception in case we're being called from another context
         logger.error(f"{e!r}")
+        raise e
     finally:
         await controller.terminate()
 
 
-async def create_did(controller):
```
michaelwiles, line 67: moved to relevant ledger or wallet facade

app/facade.py
```diff
@@ -129,145 +112,6 @@ async def post_to_ledger(url: str, payload: LedgerRequest):
     return ledger_post_res
 
 
-async def get_taa(controller):
```
michaelwiles, line 132: moved to acapy_ledger_facade

app/tests/test_wallet.py
```diff
@@ -1,9 +1,31 @@
-import routers.governance
+
```
michaelwiles, line 1: even this simple test was very helpful to validate the call was working (after the refactor)

app/acapy_wallet_facade.py
```diff
@@ -0,0 +1,84 @@
...
+    if not generate_did_res["result"]:
+        logger.error(f"Failed to create DID:\n{generate_did_res}")
+        raise HTTPException(
+            status_code=404,
+            detail=f"Something went wrong.\nCould not generate DID.\n{generate_did_res}",
+        )
+    return generate_did_res
+
+
+async def assign_public_did(controller: AriesAgentControllerBase, did: str):
```
michaelwiles, line 33: changed this method to take a did as a string (not a did object). 

The only thing used by the call is the did so did object is unnecessary abstraction.

morrieinmaas:
Overall looking really good. I need to go through this in detail again though. 

One thing that definitely needs addressing is the SonarCloud quality gate. If we introduce changes and coverage needs to be 80% for new code then this should also pass/hold true for this PR.

morrieinmaas (changes requested):

PR #25, opened 2021-06-17
Feature/9 review refactor create pub did

michaelwiles:
New pull request for refactor of create public did

Some notable changes:

1. req_header _must_ be json. Do not want to be running eval on it - that is not secure. Yes, I know that is value we control but use of eval will always be viewed suspiciously by security validation scanners. In this case it's simple to just change to using a json parser and the risk is obviated.
2. moved methods previously in facade into their relevant homes in acapy_ledger, ledger or wallet facade. Probs need some probs need more of these
3. Actually made the multi-tenant env in the docker config run as multi tenanted
4. Added web based tests - very happy that we now have tests coming in via the web and the fast api routers.
5. Added better validations. At the end a call is made to the wallet to get all the registered dids and the returned is checked that is in this list and that it is public. Be aware that because we're running tests against an existing docker we can't just check if there is one because other tests might have created another did. Once we move to a multi tenanted env there would only be one so this validation would be slightly simpler.

Subsequent update:

Was struggling with code coverage as we were not covering the error cases on the remote calls. So added tests for this. This uses an easy to use mocking framework called mockito. I find it a lot more intuitive then the more "standard" options.

Since I added this however, I discovered the error checking code was probably not working. This has been addressed. 

Though we need to do another pass as when the status codes are 400+ an aiohttp.ClientError is thrown.


.github/workflows/continuous-build.yml
```diff
@@ -21,20 +21,31 @@ jobs:
       uses: actions/setup-python@v2
       with:
         python-version: 3.9
-    - name: Install dependencies.
+    - name: Start Test Harness
       run: |
-        cp app/* . -R
-        rm -rf app
+       chmod u+x manage
+        ./manage up-as-daemon
```
michaelwiles, line 27: the new call - up as daemon is used which runs the "up" process as a daemon

app/requirements.txt
```diff
@@ -56,3 +56,4 @@ wcwidth==0.2.5
 webencodings==0.5.1
 yarl==1.6.3
 pytest-cov==2.12.1
+httpx==0.18.1
```
michaelwiles, line 59: the recommended mechanism for running async web based tests

Issue #41, opened 2021-06-28
API endpoints that that manage schemas


michaelwiles:
- [ ] Write Schema and credential definition to ledger
- [ ] write credential definition for schema id to ledger
- [ ] retrieve approved schema list from schema registry
- [ ] delete approved schema from schema registry
- [ ] retrieve approved schema from schema registry
- [ ] update approved schema in schema registry
- [ ] add approved schema to schema registry

PR #43, opened 2021-06-28
#41 API endpoints that that manage schemas

michaelwiles:
Get schema, list schema and get schema all added - with tests

BEWARE: it currently points to a _local_ von network ledger. I wanted to run my own von network when testing and I also want to do that for the CI tests... because tests generate a _lot_ of public dids.

### DIFF
--- a/app/admin/governance/schemas.py
+++ b/app/admin/governance/schemas.py
@@ -1,0 +1,50 @@
+from typing import List, Optional
+
+from aries_cloudcontroller import AriesAgentControllerBase
+from fastapi import APIRouter, Depends
+from pydantic import BaseModel
+
+from facade import create_yoma_controller
+
+router = APIRouter(prefix="/admin/governance/schemas", tags=["Schemas"])
+
+
+class SchemaDefinition(BaseModel):
+    name: str
+    version: str
+    attributes: List[str]
+
+
+@router.get("/{schema_id}")
+async def get_schema(
+    schema_id: str,
+    aries_controller: AriesAgentControllerBase = Depends(create_yoma_controller),
+):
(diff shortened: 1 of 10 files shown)


michaelwiles:
Okay - local ledger no more. Just switched to the cloud one for now - needed to work on fast api dependency context so shelving the local ledger for now.

Merged in the changes from fast api dependency branch so it is ready to roll!

michaelwiles:
Also included the credential definition end points

morrieinmaas:
@michaelwiles can you resolve the conflicts? Then we can merge this.

morrieinmaas:
* Providing the controller via Depends is great
* not needing to pass wallet_id is also great
  * hence fixing and streamlining what’s passed in the header is great
* 3 controller types is great and neccessary
However:
* The entire logic of creating a controller and deciding which one to provide is a mess
  * it’s too butchered
  * there is duplicate logic
  * await controller.terminate() in two places: agent controller and create_controller?!
* also there is no need for a factory of a factory
* there can be two methods
  * determine controller type
  * create controller
* The controller  methods  being in facade.py makes no sense to me at all especially because
  * the controller factory lives in utils
  * everything else in facade actually does sth against the aca-py agent or the ledger
* I propose those should also live in either wallet or ledger facade
  * hence facade is now partially what was criticised about utils
  * to me there’s quite some unnecessary complication in there
* there is logic all over files now that is used in one place so why not put it in one place?
* there is still core/wallet
  * we agreed to remove this file and put its contents somewhere else
  * the file has exactly one single method (create_pub_did)
  * other similar methods live in wallet_facade or ledger_facade
  * this is unnecessary and imo inconsistent/confusing
* there is no need to add another dependency (for using assert_that) 
  * Can all be done using already available assert

PR #44, opened 2021-07-01
Showing POC of using uplink to access acapy

michaelwiles:
So [uplink](https://uplink.readthedocs.io/en/stable/index.html) is a clever library for building web api clients...

It is inspired by a java project called [retrofit](https://square.github.io/retrofit/)

Included in this pull request is a demonstration of how to use uplink to access acapy.

The key thing to note is that there is a vast amount of code reuse, so now only is the less code to write, there is also less code to go wrong.

### DIFF
--- a/app/acapy_uplink_facade.py
+++ b/app/acapy_uplink_facade.py
@@ -1,0 +1,53 @@
+from typing import List
+
+from pydantic import BaseModel, Field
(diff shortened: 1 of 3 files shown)


michaelwiles:
Closing this so it doesn't stay int he list of open PR's... won't delete the the branch

PR #45, opened 2021-07-02
Acapy Agent injection review

michaelwiles:
Pull request for adjustments to agent injection


michaelwiles:
from Mo:

The entire logic of creating a controller and deciding which one to provide is a mess-
- it’s too butchered
- there is duplicate logic
    - await controller.terminate() in two places: agent controller and create_controller?!
- also there is no need for a factory of a factory
   - there can be two methods
      - determine controller type
      - create controller
- The controller  methods  being in facade.py makes no sense to me at all especially because
   - the controller factory lives in utils
   - everything else in facade actually does sth against the aca-py agent or the ledger
      - I propose those should also live in either wallet or ledger facade
      - hence facade is now partially what was criticised about utils
- to me there’s quite some unnecessary complication in there
- there is logic all over files now that is used in one place so why not put it in one place?
- there is still core/wallet
   - we agreed to remove this file and put its contents somewhere else
   - the file has exactly one single method (create_pub_did)
      - other similar methods live in wallet_facade or ledger_facade
   - this is unnecessary and imo inconsistent/confusing

michaelwiles:
> from Mo:
> 
> The entire logic of creating a controller and deciding which one to provide is a mess-
> 
> * it’s too butchered
> * there is duplicate logic
>   
>   * await controller.terminate() in two places: agent controller and create_controller?!
> * also there is no need for a factory of a factory
>   
>   * there can be two methods
>     
>     * determine controller type
>     * create controller
> * The controller  methods  being in facade.py makes no sense to me at all especially because
>   
>   * the controller factory lives in utils
>   * everything else in facade actually does sth against the aca-py agent or the ledger
>     
>     * I propose those should also live in either wallet or ledger facade
>     * hence facade is now partially what was criticised about utils
> * to me there’s quite some unnecessary complication in there
> * there is logic all over files now that is used in one place so why not put it in one place?

@morrieinmaas I'm going to take a stab at adjusting but I'm not to sure how you want it, so let me try. 

Can I maybe just create an agent_factory module and put everything in there? i.e. move the stuff that's currently in utils there as well? I don't think we'll use the methods in utils anywhere else?

> * there is still core/wallet
>   
>   * we agreed to remove this file and put its contents somewhere else
>   * the file has exactly one single method (create_pub_did)
>     
>     * other similar methods live in wallet_facade or ledger_facade
>   * this is unnecessary and imo inconsistent/confusing

@morrieinmaas In the interests of pull request narrowing would it be okay to put this removal of "old" code in another pull request?

michaelwiles:
Got no problem with that.

Still need to deal with sonar picking up the duplication though.

Will look into that.

morrieinmaas:
awesome. 

Not sure about the code smells/duplication aka how to go about it here. Seems to me like the pattern that FastAPi wants for dependencies is what it is here. Maybe we can mark it not smelly? Maybe you have a better solution?

morrieinmaas:
also let me know how we want to do this in terms of the tests. Once we have a version we're happy with we can go and change all the tests to no longer use the _controller_factory (happy to take that on).

michaelwiles:
I suspect the duplications are because of the [_controller_factory method still being present in agent_factory](https://sonarcloud.io/code?id=didx-xyz_aries-cloudapi-python&pullRequest=45&selected=didx-xyz_aries-cloudapi-python%3Aagent_factory.py&line=146)

Will know when I next push

morrieinmaas:
That'd make sense - didn't even occur to me. 🤞 then the problem basically soves itself on tidy up

app/routers/wallet.py
```diff
@@ -7,7 +7,8 @@
 from fastapi import APIRouter, Header, HTTPException, Depends
 
 from core.wallet import create_pub_did
-from facade import create_controller, yoma_agent, member_agent
+from facade import create_controller
+from agent_factory import yoma_agent, member_agent
```
morrieinmaas, line 11: can come from dependencies

app/tests/conftest.py
```diff
@@ -6,10 +6,9 @@
 from aries_cloudcontroller.controllers.wallet import WalletController
 from mockito import mock
 
-import facade
+import agent_factory
```
morrieinmaas, line 9: will then also come from dependencies so
```
import dependencies
```

app/tests/test_agent_factory.py
```diff
@@ -1,30 +1,47 @@
 import pytest
 from aries_cloudcontroller import AriesAgentController, AriesTenantController
 from fastapi import HTTPException
-import facade
-from utils import controller_factory, ControllerType
+
+import agent_factory
```
morrieinmaas, line 5: also from dependencies - no factory needed

morrieinmaas (changes requested):

app/facade.py
```diff
@@ -1,104 +1,22 @@
...
-        raise e
-    finally:
-        if controller:
-            await controller.terminate()
-
 
+# @deprecated(
+#     "moved to agent_factory - kept here because it is still in use - need to migrate"
+# )
 @asynccontextmanager
```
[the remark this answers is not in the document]
michaelwiles: I believe there's some places that still use it - though if you want to remove I'm happy with that.

PR #47, opened 2021-07-02
WIP Feature/new did endpoints

morrieinmaas:
Has to wait on #43 and  #45. 

Contains new admin dids endpoints for trusted registry and trusted partner


app/admin/governance/dids.py
```diff
@@ -0,0 +1,71 @@
...
+    -----------
+    api_key: Header(None)
+        The admin API key
+
+    Returns:
+    --------
+    trusted_partners: [dict/JSON]
+        List of unique trusted partners
+    """
+    all_dids_on_ledger = await aries_controller.wallet.get_dids()
```
michaelwiles, line 31: Is this the way we'll store trusted partners? I thought did's in a wallet are ones that belong to that entity. I guess it's nice if this can work this way.

I assumed there was a mechanism for looking up public did's on the ledger for which the yoma agent is the endorser... that was my impression of how this would work. Though it's probably better to do it by storing it in the wallet...

morrieinmaas: This list should be generated on the fly. this returns the results that are "public" 

Not sure, maybe there is a better wy of doing it (directly from the ledger?)

morrieinmaas: let me know if you can think of a better way

michaelwiles: @lohanspies please help...

app/tests/admin/governance/dids/test_dids.py
```diff
@@ -0,0 +1,44 @@
...
+    assert len(result) == 1
+
+    res_method = await get_trusted_registry(aries_controller=yoma_agent)
+    assert res_method == result
+
+
+@pytest.mark.asyncio
+async def test_get_trusted_partner(async_client, yoma_agent):
+
+    partner_id = "jhoLFYW6LfMEPUEZWR18A"  # YOMA
```
michaelwiles, line 29: Tests need to run as independently as possible and thus where possible not not depend on some data already being present. Tests need to be built as robust as possible.

We also cannot depend on the order of tests being predictable.

There is at least [existing code for adding a public did in the test context](https://github.com/didx-xyz/aries-cloudapi-python/blob/d2280a173bbe383c9b5988b59b5bbff607bc4d7a/app/tests/admin/governance/schemas/test_schemas.py#L96). Possible be worth extracting that into a fixture.

app/admin/governance/dids.py
morrieinmaas: I discussed this yesterday with @lohanspies - thought this was fine. But if we can make this better, please, say @lohanspies

app/tests/admin/governance/dids/test_dids.py
morrieinmaas: agreed!

morrieinmaas: any suggestions for this specific case?

morrieinmaas: @michaelwiles

morrieinmaas:
@michaelwiles fixed the test. let me know your thoughts.

This needs to wait for #45 and/or #43 because it uses dependency injection which is not on the development branch yet

Issue #48, opened 2021-07-05
create connections endpoints


morrieinmaas:
Create endpoints to establish connections between peers based on insomnia specs and new app arch.

mrcbond88:
- [] Create tests for endpoints

PR #49, opened 2021-07-05
Connection endpoints

morrieinmaas:
Scaffolding of desired connection endpoints.

closing #48


morrieinmaas:
@vineeth14 please have a look/do a code review when you find the time

morrieinmaas:
@vineeth14 I can run the tests locally without a problem. Not sure yet why they fail in GH actions. I'll do some digging.

morrieinmaas:
@vineeth14 ready for review 👍🤗

app/generic/connections.py
```diff
@@ -0,0 +1,54 @@
...
+from aries_cloudcontroller import AriesAgentControllerBase
+
+from dependencies import member_agent
+
+logger = logging.getLogger(__name__)
+
+
+router = APIRouter(prefix="/generic/connections", tags=["connections"])
+
+
```
vineeth14, line 13: Should these calls be limited to only the member_agent?

morrieinmaas: hmm I was wondering the same. Should cover most cases though imo. In the end everything you wanna connect to should be a sub wallet (except for YOMA wallet itself maybe?). My take on this is we can always create these endpoints for/extend them to whatever extra case we have. This should be the general use case though. That said if you think I'm wrong here or missing something please do say

morrieinmaas:
@vineeth14 you can merge it and make it generic afetr #58 is merged imo

morrieinmaas:
@vineeth14 Please have a look and merge when ready

vineeth14:
Cool, will have a look

PR #54, opened 2021-07-08
Additional Wallet Endpoints 

### LINKED ISSUE #51: Create wallet endpoints
Wallet endpoints to be created: (everything in AcaPy)
- [x] list all dids
- [x] create local did
- [x] rotate-keypair
- [x] assign public did

vineeth14:
ToDO
Wallet- related Endpoints
Testing

closes #51


vineeth14:
ToDO
Wallet- related Endpoints
Testing

closes #51

michaelwiles:
These are not just for the yoma agent right? These are end points that are used to manage _my_ wallet. i.e a members, ecosystem or yoma agents wallet.

Thus I'm... 
* not sure the endpoint should be in the admin context, endpoint should just be prefixed by /wallet (as it is in acapy) 
* endpoint needs to have the fast api dependency that selects the correct agent based on the role provided - what can be done at this stage is to provide a place holder selecting function (suggestion is to call this all_agents) which for now only returns say the member agent or to make it easier the yoma agent while we wait for that selector function to be finalised.

PR #58, opened 2021-07-09
Selecting agent based on role

michaelwiles:
So this is a poc at this stage.

It uses the x_role header to switch to the relevant agent.

So not sure if there are existing end points (in development branch) that use a token to authenticate so not sure if I can actually use this new fast api depends function. But suffice to say when the end point is shared the following will be used...

```
@router.get("/shared-endoint", tags=["did"])
async def get_api_call_that_works_on_both_member_and_ecosystem(
    aries_controller: AriesAgentControllerBase = Depends(ecosystem_or_member_agent),
):

```
see other comments inline

UPDATE: promoted from poc - now becomes the real code

### DIFF
--- a/app/dependencies.py
+++ b/app/dependencies.py
@@ -1,3 +1,5 @@
+from contextlib import asynccontextmanager
+
 from fastapi import Header
 
 import logging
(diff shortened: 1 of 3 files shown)


app/dependencies.py
```diff
@@ -39,6 +41,24 @@ async def yoma_agent(x_api_key: str = Header(None)):
             await agent.terminate()
 
 
+async def ecosystem_or_member_agent(
```
michaelwiles, line 44: Not sure if there are any existing end points that require an admin version of the eco system agent or member agent. If there are an equivalent method can be added that creates the "admin" version (admin = multitenant end points added).

Can add a method cann admin_acosystem_or_admin_member_agent and then inject that via  depends where relavant

michaelwiles:
forgot to restore the changed urls... will do so asap

michaelwiles:
* promote to real
* change dependency functions
* change to agent selector
* add admin agent selector
* fix tests where relevant

app/dependencies.py
michaelwiles: Have added agent_selector and admin_agent_selector and full tests

Issue #57, opened 2021-07-09
Update the open api definition (swagger) for our api and publish it on insomnia


mrcbond88:
Define interface for clients to be able to call an API.

This will also give us a clear definition of what the API does.

michaelwiles:
Have started this. clone the cloud api repo into insomnia and you'll get some results.

For help on doing this:

To add it to insomnia you need to do the following:
1. open designer, click create (purpose button top right), click clone
2. enter github location of cloudapi (https) - enter email, name and username (github username)
3. you'll need to get a personal access token - follow the instructions in the link to generate the access token
4. This will then fetch the current version

Am going to cleanup the tags etc to make the generated openapi spec more useful.

michaelwiles:
I've done some work on this - clone the cloud api repo in insomnia and you'll see what's happening

michaelwiles:
I'm going to work a little on fixing the tags and making the generated yaml better.

PR #62, opened 2021-07-13
#57 Enriched the open api definition that is published and also published version in Yaml

michaelwiles:
Can now go to /openapi.yaml to render the open api definition in yaml.

But got of /docs to see it in the nice html view.


app/admin/governance/multitenant_wallet/wallet.py
```diff
@@ -12,10 +12,10 @@
 
 logger = logging.getLogger(__name__)
 
-router = APIRouter(prefix="/wallets", tags=["wallets"])
+router = APIRouter(prefix="/wallets", tags=["Admin: Wallets"])
```
wip-abramson, line 15: Should this not be /admin/wallets perhaps

michaelwiles: Thanks for your input there @wip-abramson - I do agree. Though it's not part of this PR's purview.

PR #69, opened 2021-07-15
added error handler for dealing elegantly with ClientResponseError

michaelwiles:
ClientResponseError is what is sent back by the arise cloud controller.

Currently we just return 500 when we get an error in cloud controller as we don't handle this exception. This change will modify things so that we do handle this exception elegantly.


michaelwiles:
Merging - that codacy issue is challenging to remove

PR #75, opened 2021-07-19
Issue/74/fixtures per agent

michaelwiles:
PR for the new async client agents for bob and alice.

What is awesome is that the headers can be set when creating the async client so as long as you use the correct async client you will run as the relevant agent.

The wallet gets deleted after it has been created as well as well as the connections that are created.


morrieinmaas:
Closing as stale

PR #77, opened 2021-07-20
Switching off ngrok and superfluous images

michaelwiles:
This pull request removes the requirement for ngrok - it doens't completed remove the configuration - just removes the references to it.

Key Changes:
1. the removal of the ngrok images from the manage script - thus the ngrok images are not started
2. The adjustment to the config to set the AGENT_ENDPOINT to the location of the relevant images in the docker network 
   _this_ is the money change. The endpoint used to be the one generated by ngrok. With this change we do docker based coms instead of going via ngrok.
3. Some housekeeping changes: remove the script that waits for the ngrok images to be up and the tunnel functional.


dockerfiles/agents/Dockerfile.agent
```diff
@@ -18,4 +18,4 @@ RUN chmod +x ./wait.sh
 
 USER $user
 
-CMD ./wait.sh ./startup.sh
```
michaelwiles, line 21: removed the wait - this is actually ngrok-wait.sh as it's renamed when copied onto the docker. No longer required as ngrok is disabled

environments/yoma-ga/aca-py-agent.default.env
```diff
@@ -59,7 +59,7 @@ ACAPY_GENESIS_URL=http://testnet.didx.xyz/genesis
 
 
 # Name of ngrok container if exposing agent endpoint over ngrok
-NGROK_NAME=ngrok-yoma-ga
```
michaelwiles, line 62: ngrok container not required

environments/yoma-ga/aca-py-agent.default.env
```diff
@@ -75,4 +75,6 @@ ACAPY_AUTO_RESPOND_PRESENTATION_PROPOSAL=false
 ACAPY_AUTO_RESPOND_PRESENTATION_REQUEST=false
 ACAPY_AUTO_STORE_CREDENTIAL=false
 ACAPY_AUTO_VERIFY_PRESENTATION=false
-ACAPY_PRESERVE_EXCHANGE_RECORDS=true
\ No newline at end of file
+ACAPY_PRESERVE_EXCHANGE_RECORDS=true
+
+AGENT_ENDPOINT=http://yoma-ga-agent:3020
```
michaelwiles, line 80: this is the important big. The AGENT_ENDPOINT is what an agent uses to provide connections to itself. This used to be generated by the ngrok startup. 

The ngrok wait script used to set this value so maybe to plugin and plugout ngrok we just enable/disable the ngrok wait script - by copying a blank or copying the real thing... and the real thing will override this env var...

manage
```diff
@@ -54,9 +54,8 @@ EOF
 # -----------------------------------------------------------------------------------------------------------------
 # Default Settings:
 # -----------------------------------------------------------------------------------------------------------------
-DEFAULT_CONTAINERS="yoma-multitenant-wallets-db yoma-multitenant-web ngrok-yoma-multitenant yoma-multitenant-agent
-                    yoma-ga-wallets-db yoma-ga-web ngrok-yoma-ga yoma-ga-agent"
-                    # tails-server"
+DEFAULT_CONTAINERS="yoma-multitenant-wallets-db yoma-multitenant-agent
```
michaelwiles, line 57: notice how only the acapy agents are running. This is all that is actually required as the fastapi runs by itself when running tests - i.e. the tests start it.

I'd prefer to not have the fastapi running at all as we can just start fastapi via a fast api startup script and then also allow debugging of it. Debugging is challenging when running something on docker - or at least more challenging when running something natively.

michaelwiles: fastapi can be run with:

`uvicon main:app --reload --port 8082`

We get auto reload and we get a custom port and then if we run this in debug mode we can set break points...

scripts/startup.sh
```diff
@@ -1,3 +1,4 @@
+#!/bin/bash
```
michaelwiles, line 1: necessary as the ngrok-wait script was setting the shell before

PR #80, opened 2021-07-26
CloudController Refresh: Using uplink generated client interfaces

michaelwiles:
So I have:

1. Generated the full api with uplink - Not every servce is ready as it needs some additional elements.
2. Provided as the file api.py - this can be moved to cloud controller
3. Modified the code for wallet to use this
New uplink contains ALL schemas for all calls so needed to move from a json approach to an object approach
4. Changes made to dependencies so no other changes required - AriesAgentController is still used except that the old wallet object is replaced by a a new one.


michaelwiles:
Apologies for all the failures... will get to them

morrieinmaas:
Sorted out in #98

Issue #86, opened 2021-07-28
Standardize  Tags


vineeth14:
Decide on the format for the existing endpoint's tags.
Ensure the changes enforced are consistent.

vineeth14:
Going with a lowercase convention for tags.
Admin tags have the following format- "admin:"

PR #88, opened 2021-07-30
Enhancement/86/standard tags

vineeth14:
#86


app/admin/governance/multitenant_wallet/wallet_admin.py
```diff
@@ -15,7 +15,7 @@
 
 logger = logging.getLogger(__name__)
 
-router = APIRouter(prefix="/admin/wallet-multitenant", tags=["Admin: Wallets"])
+router = APIRouter(prefix="/admin/wallet-multitenant", tags=["admin: wallet"])
```
morrieinmaas, line 19: I'd say either we do wallets and schemas or wallet and schema? Or did we decide to use singular for one and plural for the other?

app/admin/governance/schemas.py
```diff
@@ -6,7 +6,7 @@
 
 from dependencies import yoma_agent
 
-router = APIRouter(prefix="/admin/governance/schemas", tags=["Admin: Schemas"])
+router = APIRouter(prefix="/admin/governance/schemas", tags=["admin: schemas"])
```
morrieinmaas, line 9: see previous comment

morrieinmaas (changes requested):
Hey nice work overall. Just not sure about singular and plural eg at schemas and wallet(s). Maybe that is on purpose and I missed that information? Let's quickly clarify that (or change it). Happy to merge then

PR #89, opened 2021-08-05
Feature/local von

vineeth14:
Added Local Von feature to aries-cloudapi


morrieinmaas:
The problem with the SSL error on GH actions build is now fixed. It appears that the problem is in docker-compose (version). After some digging and trying various things [this](https://github.com/docker/compose/issues/7686) post's suggestions to install docker-compose via pip fixes the issue. 

However, now there is 24 tests failing. @vineeth14 you said they are all working fine for you locally? Can you have a look? They also fail for me locally just as they do on GH actions. Maybe you have an idea?

PR #97, opened 2021-08-23
Enhancement/96/docstrings

vineeth14:


app/generic/connections.py
```diff
@@ -23,6 +26,9 @@ async def accept_invite(
     invite: dict,
     aries_controller: AriesAgentControllerBase = Depends(agent_selector),
 ):
+    """
+    Accept connection invite.
```
morrieinmaas, line 30: Param is the invite object

app/generic/connections.py
```diff
@@ -40,6 +49,14 @@ async def get_connection_by_id(
     connection_id: str,
     aries_controller: AriesAgentControllerBase = Depends(agent_selector),
 ):
+    """
+    Retrieve connection by id.
+
+    Parameters:
+    -----------
+    connection_id: str
+        connection id
```
morrieinmaas, line 58: to me this is clear without the extra connection id descriptio line

app/generic/connections.py
```diff
@@ -49,5 +66,12 @@ async def delete_connection_by_id(
     connection_id: str,
     aries_controller: AriesAgentControllerBase = Depends(agent_selector),
 ):
+    """
+    Delete connection by id.
+
+    Parameters:
+    -----------
+    connection_id: str
```
morrieinmaas, line 74: this is plenty readable to me (see comment above)

app/generic/issuer_v2.py
```diff
@@ -61,6 +74,9 @@ async def get_records(
     connection_id: Optional[str] = Header(None),
     aries_controller: AriesAgentControllerBase = Depends(agent_selector),
 ):
+    """
+    Get list of records.
+    """
```
morrieinmaas, line 79: optional connection id param here

app/generic/issuer_v2.py
```diff
@@ -86,6 +114,13 @@ async def get_credential(
     credential_id=Header(...),
     aries_controller: AriesAgentControllerBase = Depends(agent_selector),
 ):
+    """
+    Get credential by credential id
+    Parameters:
+    -----------
+    credential_id: str
+        credential identifier
+    """
```
morrieinmaas, line 123: returns?

app/admin/governance/credential_definitions.py
```diff
@@ -22,6 +22,18 @@ async def create_credential_definition(
...
+    Create a credential definition.
+
+    Parameters:
+    -----------
+    credential_definition: CredentialDefinition
+        Payload for creating a credential definition.
+
+    Returns:
+    --------
+    The response object obtained from creating a credential definition.
```
morrieinmaas, line 35: The following applies to much of the comments below also:

Do we want to either at least give a pydantic model for return type (when python standard type is not applicable) or do we possibly even want to provide n example of the return objects structure). For example, should we return an invitation should we ensure we have an invitation pydantic model (or other custom type for that) we reference here and a developer can look up or should be reference it it here AND even provide a a description of the strcuture/fields?

morrieinmaas (changes requested):
Hey good effort and thanks for taking this on. There is two things I'd go over before merging. 

1. Ensure the structure is always coherent. 
    1. Description of what the function does
    2. input params
    3. return value/type
 2. Ensure return type is of python standard type or pydantic model (for the latter possibly provide the structure?; not 100% sure about that because that could also make this pretty bloaty and annoying to read the actual code)

Issue #59, opened 2021-07-09
Add class schamas to the existing calls where appropriate


michaelwiles:
So I'd like us to avoid using dicts as input. 

The structure of the data that we received must be made explicit in our code base and hopefully, since fastapi and pydantic play nicely together this will be a simple process.

As dev stands now the multi teannt subwallet calls need these schema classes added. It's not a lot of work. These classes will just be added to the existing wallet module in multi_tenant.

Issue #13, opened 2021-06-08
Replace endpoint return types json/dict with pydantic schemas 


morrieinmaas:
We can use pydentic models (correlated to the JSON schemas in FastAPI, [see also](https://fastapi.tiangolo.com/tutorial/response-model/)) to replace the rather undefined json returns for as many endpoints as possible. 

1. That'll enable us to provide function signatures for the endpoints
2. We have a better handle on/it is more clear what we have to expect when calling a function
3. We get validation of data structures coherently and for a discount. 
4. We can test this more easily
5. Some other benefits...

Downsides: ? - Please leave them in the comments if you find any.

Happy to work on this. It's more a matter of time/priority. So, it's up for grabs - first come first serve.

vineeth14:
Good Idea! Will also keep this in mind while developing the verifier endpoint.

PR #98, opened 2021-08-24
move to uplink acapy client

TimoGlastra:
Signed-off-by: Timo Glastra <timo@animo.id>

- [x] Fix tests for those which aren't working
- [x] Fix multi-tennant endpoints
- [x] Fix pubDID
- [x] release cloudcontroller "final" version (working for what we have) and pull that in requirements via pypi instead of from a git branch/commit


app/generic/connections.py
```diff
@@ -10,37 +10,38 @@
...
+    invite = await aries_controller.connection.create_invitation()
     return invite
 
 
-@router.post("/accept-invite")
+@router.post("/accept-invite", response_model=ConnRecord)
 async def accept_invite(
-    invite: dict,
-    aries_controller: AriesAgentControllerBase = Depends(agent_selector),
+    invite: ReceiveInvitationRequest,
```
TimoGlastra, line 26: Input models can also be specified based on the cloud controller models

TimoGlastra (commented):
I've updated the connections endpoints to use the generated cloud controller and added some comments for context.

TimoGlastra:
I've updated the cloud controller to 0.3.0a7, which contains some fixes:
- Updated the OpenAPI with some manual fixes. I'll see later to get those into ACA-Py itself (mainly ledger operations)
- Added support for anyOf / oneOf models

TimoGlastra:
@morrieinmaas @vineeth14 Have you run into any issues with using the cloud controller?

morrieinmaas (approved):

morrieinmaas:
@TimoGlastra I made the tests pass now and removed some code smells. There is one test that needs fixing which is `test_send_credential_proposal` (for that please see [this issue in the controller](https://github.com/didx-xyz/aries-cloudcontroller-python/issues/62)). 

Please also have a look at issues 55-61 in the cloudcontroller. I've tweaked the controller for those issue on am now installing the controller from a branch via git+https://github.com/didx-xyz/aries-cloudcontroller-python@1162bb8b51ef5ddcaec916708c66966e3f18c358 as dependency. Before we we fully move to uplink version of the controller I'd say we should fix the issues and return types. That said, as you'll see in the code replacing a PyDantic model with an ANy type returns a couroutine object instead of a model object so I had to await it's JSOn response seprately. That is not ideal either, but works for now and will disappear when types are fixed. 

Do NOT merge yet.

morrieinmaas:
closes #13 and closes #59

PR #114, opened 2021-09-26
feat: issuer

### LINKED ISSUE #102: Issue credentials
OPs must be able to issue credentials to youth upon successful completion of opportunity.

We must work towards three end points: issue, accept and decline (do nothing for decline). The rest must be automated.

- [ ] Send credential offer that specifies the credential. Specify the definition, attributes and connection ID

- [ ] Member able to view and accept credential. Accept as is.

- [ ] Yoma member able to decline/reject credential. No API endpoint is called in this instance. This must be kept in the notification list. The member must be able to accept the credential at a later stage, or be able to clear the notifications/credential.

TimoGlastra:
Initial implementation for issuer. Still needs fine tuning and tests, but wanted to get some initial feedback.

Closes #102

This implementation combines the already existing issuer v1 and issuer v2 into a single API that can handle both. The idea behind it, is that as an user of the cloudapi, you don't want to concern with the specifics of the protocol version you're using. The only thing you need to know is which version to use (when initiating).

When issuing a credential you specify which version you'd like to use.

```
POST /generic/issuer/credentials
```
```json
{
  "connection_id": "b93dae4c-90c5-4159-b3b5-4dcdc47e7272",
  "cred_def_id": "ExC9DD7C1P4axwRM5EiHys:3:CL:8:test",
  "attributes": {
    "string": "test"
  },
  "protocol_version": "v2"
}
```

The call is the same except for the difference in "v1" vs "v2". To accomodate for the difference in data model between the two protocol versions a simplified model has been made that is used as record model.

Whenever a record representation is returned (either v1 or v2) it will be in the following format. More fields can be added as needed

```json
{
  "credential_id": "v2-5158e9c9-a2d7-4582-90f8-e692241ad3a2",
  "role": "issuer",
  "created_at": "2021-09-26 14:40:46.160974Z",
  "updated_at": "2021-09-26 14:40:46.160974Z",
  "attributes": {
    "string": "test"
  },
  "protocol_version": "v2",
  "schema_id": "ExC9DD7C1P4axwRM5EiHys:2:test:1.0.0",
  "credential_definition_id": "ExC9DD7C1P4axwRM5EiHys:3:CL:8:test",
  "state": "offer-sent",
  "connection_id": "b93dae4c-90c5-4159-b3b5-4dcdc47e7272"
}
```

I think one of the benefits of this approach is that we can abstract the complexity of dealing with multiple versions of protocols. Now just for issue credential, but the same could be applied for present proof (v1, v2) and connections (didexchange, connections).

As you can see in the `credential_id` property, the id is prefixed with `v2-`. For v1 it will be prefixed with `v1-`. When calling the api, this is what allows the cloudapi to know whether it should call the v1 or v2 endpoints. One of the caveats of this approach, is that we are now required to also process the webhooks as just an id (without the `v1-` or `v2-` prefix) is not a valid id anymore for the cloudapi. So before sending the plain webhook event, we should include the cloudapi id somewhere.

I'd like to know if this is a valid tradeoff?

Another approach would be for every call to include the version explicitly, so we can keep the plain identifier. But I think it would be good to create a common data model for the webhook payloads either way, as otherwise the use has to deal with the differences in structure for v1 and v2 records. I think it would be nice if we can completely abstract this.

Please let me know what you think


app/generic/issuer/facades/acapy_issuer_v1.py
```diff
@@ -0,0 +1,146 @@
...
+        record = await controller.issue_credential_v1_0.issue_credential_automated(
+            body=V10CredentialProposalRequestMand(
+                connection_id=credential.connection_id,
+                credential_proposal=credential_preview,
+                auto_remove=False,
+                cred_def_id=credential.cred_def_id,
+            )
+        )
+
+        return IssuerV1.__record_to_model(record)
```
morrieinmaas, line 68: should this maybe be `cls.__record_to_model(record)`?

And corallary below?

app/generic/issuer/facades/acapy_issuer_v1.py
```diff
@@ -0,0 +1,146 @@
+import logging
```
morrieinmaas, line 1: genereally, but that's really personal style/preference, why make this a class/classes at all? Maybe functions is more lightweight/faster? You could access it like a class via import eg
```python
import .generic.issuer.facade.acapy_issuer_v1 as IssuerV1
```

then call the methods on that which are just functions. That saves instantiating all that dundar method stuff one gets for class objects.

really just as an idea. Totally fine if you wanna do it this way.

morrieinmaas (commented):
Nice work and thanks for putting in all the ffort🙏

Generally:

1. What about tests?
2. cls instead of class namespace? see comment in changes
3. Maybe not make this classes (see comments). But really more a style/preference question

Cheers

morrieinmaas:
Hey, that was fast.

1. Awesome
2. hmm well maybe I'm missing somethign or maybe the typechecker/linter is wrong here. Might be me though. I'm gonna have another look into cls/classmethod decorator and check. One argument why I thought it made sense using cls is that it's for the class (not instance) and so if one inherits from or composes with the IssuerV2 class that is still valid and not bound to the namespace.
3. Thought so kinda. Really more an idea. Your way is equally valid jsut slightly different. I see (dis)advantages in doing either. Do as you please.

TimoGlastra:
Re 2. I think you're right. It makes sense what you're saying. I just checked again and it is not giving that error, so it may have been a derp on my side 🤷

TimoGlastra:
Two more questions :)

1. What do you think of the combining of the credential ids? The public API only has a single credential id, and it is used to manage both the credential, and the credential exchange record. If you delete the one, the other will also be removed. This is to remove some of the complexities around credentials in ACA-Py.
2. What do you think of the `v1-` and `v2-` prefix to the credential id to distinguish between the protocol version? As this can have an impact on other components (such as the webhooks) I'd like to make sure you also think this is a good idea.

morrieinmaas:
1. If I understand that correctly - 👍.
2. I think that's quite thoughtful. However, isn't that described via `protocol_version` attr? I sort of agree with your reasoning/why you do it. I'm just a little hestitant because if we create eg CredentialExchange objects and the credential_id has 'v1-' or 'v2-' prefix will that work passing it to aca-py or will aca-py throw an error for invalid id (format)?

TimoGlastra:
> 2\. I think that's quite thoughtful. However, isn't that described via `protocol_version` attr? I sort of agree with your reasoning/why you do it. I'm just a little hestitant because if we create eg CredentialExchange objects and the credential_id has 'v1-' or 'v2-' prefix will that work passing it to aca-py or will aca-py throw an error for invalid id (format)?

Yeah I get why you're hesitant. protocol_version is passed when issuing a credential. But for the other endpoints where you provide an id, it needs to know which version to use. Basically the unique identifier for a record is now the cred_ex_id combined with the protocol version. I could also extract the protocol version from the id and make it a separate parameter for every call, but we need both. Then it'd be up to the caller to specify which version a specific ID belongs to. I thought it may be nice if we could abstract it completely. It won't work with ACA-Py directly, you do have to remove the prefix before passing the ID to ACA-Py.

morrieinmaas:
> > 2. I think that's quite thoughtful. However, isn't that described via `protocol_version` attr? I sort of agree with your reasoning/why you do it. I'm just a little hestitant because if we create eg CredentialExchange objects and the credential_id has 'v1-' or 'v2-' prefix will that work passing it to aca-py or will aca-py throw an error for invalid id (format)?
> 
> Yeah I get why you're hesitant. protocol_version is passed when issuing a credential. But for the other endpoints where you provide an id, it needs to know which version to use. Basically the unique identifier for a record is now the cred_ex_id combined with the protocol version. I could also extract the protocol version from the id and make it a separate parameter for every call, but we need both. Then it'd be up to the caller to specify which version a specific ID belongs to. I thought it may be nice if we could abstract it completely. It won't work with ACA-Py directly, you do have to remove the prefix before passing the ID to ACA-Py.

Alright I see. [**phrase**](https://i.redd.it/bhu14ykms2241.jpg),  I'd say go for it. Only worried that that might trip people up having to remove the rpefix to in order to pass it to aca-py - I'm a bit on the fence here. Sicne it's really only codebase internal and the future client won't have to deal with it, go for it. I mean more abstraction is what we said we're aiming for so ...

morrieinmaas:
@TimoGlastra That looks fab 🚀.

app/requirements.txt
```diff
@@ -27,7 +27,7 @@ mypy-extensions==0.4.3
 packaging==20.9
 pathspec==0.8.1
 pep517==0.8.2
-Pillow==8.2.0
+Pillow==8.3.2
```
morrieinmaas, line 30: Nice - did the same thing also ;)

morrieinmaas (approved):
Looks great 🔥

Issue #103, opened 2021-08-25
Request proof


mrcbond88:
We can use string matching to find a specific string in a credential. 

- [ ] Request proof from a connection ID
- [ ] Create proof request
- [ ] Accept proof request - how do you want to satisfy the proof request
- [ ] Decline proof request - we can remove the presentation exchange record from the wallet

If a request is sent & there is no credential, then you can't prove anything and there's nothing to send back.

- [ ] Unify API between v1 and v2
- [ ] Add Integration tests
- [ ] Add endpoint to retrieve credentials for a proof request

--- Not for V1 ---
Once we have a rating system, we can use predicates to assess level of skill.

- [ ] Filter options on how you can satisfy a proof request

morrieinmaas:
TODO:

[ ] Add retrieve credentials for a requetsed proof

returns list of available credentials based on /present=-proof/credentails{cred_ex_id}/credentials

TimoGlastra:
- [ ] Unify API between v1 and v2
- [ ] Add Integration tests
- [ ] Add endpoint to retrieve credentials for a proof request

TimoGlastra:
@morrieinmaas could you also take a look at https://github.com/didx-xyz/aries-cloudapi-python/issues/129 as part of this PR? So disabling proofs auto-accept where applicable

Issue #99, opened 2021-08-25
Ecosystem management


mrcbond88:
## TODO

- [ ] #190
- [x] https://github.com/didx-xyz/aries-cloudapi-python/issues/176
- [x] https://github.com/didx-xyz/aries-cloudapi-python/issues/159
- [x] #157
- [x] https://github.com/didx-xyz/aries-cloudapi-python/issues/186
- [x] https://github.com/didx-xyz/aries-cloudapi-python/issues/176
- [x] https://github.com/didx-xyz/aries-cloudapi-python/issues/185
- [x] https://github.com/didx-xyz/aries-cloudapi-python/issues/141
- [x] https://github.com/didx-xyz/aries-cloudapi-python/issues/143
- [x] https://github.com/didx-xyz/aries-cloudapi-python/issues/144

## Requirements

Required from API to spin up ecosystem partners:

- For issuers and verifiers
	- Create subwallet for ecosystem partners to be able to act as issuer, verifiers & members.
- For verifiers 
  - Add verifiers to trust registry in the role of verifier with enough context to make establish connections with/from yoma members.
- For issuers 
	  - When a partner onboards as an issuer, Yoma will automatically write a public DID for them.
	  - Provide schema list to ecosystem partners( as role of issuer). The endpoint exists and will call the schema list for partners to choose from.
	  - Once schema is chosen, allow ecosystem subwallets to write to ledger via transaction endorser protocol. (Only credential definitions)
- For all
	- Enforce the trust registry rules

mrcbond88:
Working through Timo's endpoints, tests and create pubDID to see which ones are outstanding and which ones are not working. From there once we've got our ducks in a row, we'll be a lot quicker to move through this.

PR #123, opened 2021-10-15
[WIP] Proposal for convinience facede to registry.

### LINKED ISSUE #122: Create trust registry facade
#113 has introduced a simple trust regsitry where schemas and partners (trustees) of the ecosystem will be stored.

#103 and #99 both require checking information about schemas and partner against that registry in order to validate whether a certain requested action requested via an endpoint is valid to perform.

Further details are in #99 #103 respectively.

Here, we should aim for providing convenient access to the registry e.g. via a 'facade'. The PR should provide convenience wrappers for commonly used/required validation against the registry such as checking whetehr a given partner has the appropriate role attribute ("issuer", "verifier", both), whether a did is public, obtaining DID from invite (what to do with multi-use invites), or whether a schema is valid/registered.

morrieinmaas:
closes #122 

@TimoGlastra here's a quick proposal for what I meant in #122. Maybe that'll take some of the cumbersome bits away from #99 and #103. If you have something else/better in mind, just say. Happy to approach it another way.


app/trust_registry_facade.py
```diff
@@ -0,0 +1,57 @@
...
+
+import requests
+
+TRUST_REGISTRY_URL = os.getenv("TRUST_REGISTRY_URL", "http://localhost:8001/")
+
+logger = logging.getLogger(__name__)
+
+
+async def actor_has_role(
+    actor_id: str, role: str = Union[Literal["issuer"], Literal["verifier"]]
```
TimoGlastra, line 14: I think you can just do:
```suggestion
    actor_id: str, role: str = Literal["issuer", "verifier"]
```

TimoGlastra: Also maybe we can make one role type and import that everywhere to make it easier to add roles over time?

```
Role = Literal["issuer", "verifier"]
```

app/trust_registry_facade.py
```diff
@@ -0,0 +1,57 @@
...
+    actors = requests.get(TRUST_REGISTRY_URL + "/registry/actors")
+    actors_with_role_list = []
+    if actors.status_code != 200:
+        return actors_with_role_list
+    [
+        actors_with_role_list.append(actor)
+        for actor in actors.json()["actors"]
+        if role in actor["roles"]
+    ]
+    return actors_with_role_list
```
TimoGlastra, line 34: This feels a bit weird. The array blocks are needed I guess, but we're not using it. This would make more sense to me:

```suggestion
    if actors.status_code != 200:
        return []
        
    actors_with_role_list = [
    	actor for actor in actors.json()["actors"]
        if role in actor["roles"]
    ]
    return actors_with_role_list
```

app/trust_registry_facade.py
```diff
@@ -0,0 +1,57 @@
...
+    return bool(role in actor_res.json()["roles"])
+
+
+async def actors_with_role(
+    role: str = Union[Literal["issuer"], Literal["verifier"]]
+) -> list:
+    actors = requests.get(TRUST_REGISTRY_URL + "/registry/actors")
+    actors_with_role_list = []
+    if actors.status_code != 200:
+        return actors_with_role_list
```
TimoGlastra, line 28: Not sure if we should just swallow errors

morrieinmaas:
> Thanks for this. I like the approach so If you could work this out in more detail that would be great.
> 
> One note is that it feels like a bit of overhead to call the trust registry api from within the trust registry api. I.e. couldn't we just directly call the internals of the trust registry?
> 
> Great job 👍

Thanks for the quick feedback. I'm happy with pretty much all the code change suggestions you made. 

Regarding:
> One note is that it feels like a bit of overhead to call the trust registry api from within the trust registry api. I.e. couldn't we just directly call the internals of the trust registry?
Not entirely sure what you mean here? As previously suggested, the other option that comes to mind is to provide such functionality from the trustregistry itself. I.e. the trust registry would have endpoints that return answers to those questions such as does actor x have role a, b, both, or None. Effictively that would move that logic from the cloudapi to the trust registry. I could also implement that. Is that what you meant? Or something else? 

Cheers

TimoGlastra:
I'm a doofus, I thought those facade methods were new endpoints in the trust registry. But I see now it are just methods in the cloud api 🤦

morrieinmaas:
> I'm a doofus, I thought those facade methods were new endpoints in the trust registry. But I see now it are just methods in the cloud api 🤦

🤣 - no worries. Glad I'm not the only one these things happen to every now and again

coverage.xml
```diff
@@ -0,0 +1,478 @@
+<?xml version="1.0" ?>
```
TimoGlastra, line 1: I think we don't want this in the repo?

TimoGlastra:
Damn you're quick 😁

PR #125, opened 2021-10-17
enforce trust registry

TimoGlastra:
Enforce the trust registry for issuer (`assert_valid_issuer`), and also add assertion methods to be used in the proof api (`assert_valid_verifier`).

Couple notes:
- @morrieinmaas the import paths in the trust registry are giving me errors. When a relative import is done it will succeed in the tests, but in the ./manage docker containers it can't find the relative path with the following error. Any idea what it could be?
```
yoma-trust-registry_1          |   File "/app/./main.py", line 4, in <module>
yoma-trust-registry_1          |     from .registry import registry_actors, registry_schemas
yoma-trust-registry_1          | ImportError: attempted relative import with no known parent package
```
- I've added the `assert_valid_verifier` method which can verify if the verifier is valid. It takes a fully qualified did (`did:sov:xxx` or `did:key:xxx`) and verifies whether that did is registered. This means that in the view of the trust registry all DIDs are public. We just map the invitation key to a `did:key` locally before checking the trust registry. (can transform using `ed25519_verkey_to_did_key` in utils). The procedure for the verifier / prover is updated in #99
- I've updated the trust registry transformation of the roles from str <-> list into a custom sqlalchemy type. This way we can have proper validation in the pydantic scheme

Still need to update the issuer e2e tests.


morrieinmaas:
@TimoGlastra I'll have a look at that. My bad it looks. Will have a proper look tomorrow. Tests work for me though.

morrieinmaas:
@TimoGlastra Should have fixed it. Tests pass still and error is gone. Was an import issue. Let me know whether or not that works for you after pull develop and merge into this branch.

morrieinmaas:
let me have a look. Not sure from the top of my head

app/facades/trust_registry.py
```diff
@@ -1,25 +1,120 @@
...
+TRUST_REGISTRY_URL = os.getenv("TRUST_REGISTRY_URL", "http://localhost:8001")
 
 logger = logging.getLogger(__name__)
 
 Role = Literal["issuer", "verifier"]
 
 
+class TrustRegistryException(Exception):
+    """Class that represents a trust registry error"""
+
```
morrieinmaas, line 17: If you don't have any extra logic here that you want, maybe just adding `pass` makes it clearer.

app/facades/trust_registry.py
```diff
@@ -1,25 +1,120 @@
...
     if actor_res.status_code != 200:
         raise HTTPException(404, detail="Actor does not exist")
     return bool(role in actor_res.json()["roles"])
 
 
-async def actors_with_role(role: str = Role) -> list:
+async def actor_by_did(did: str) -> Optional[Actor]:
+    actor_res = requests.get(TRUST_REGISTRY_URL + f"/registry/actors/did/{did}")
+
+    if actor_res.status_code != 200:
```
morrieinmaas, line 111: This is nice and concise. I'd maybe say though that it's worth it being a bit more precise here. Say for instance the TR return 404 not found, then returning none is fine. There is also a chance that the TR is down or busy or conneciton dropped or whatever.

I'd check what exactly one gets for /registry/actors/did/{did} with did non existent and return none only for that case and raise an exception otherwise. That's simply to avoid confusion when using it and the user/program is not false tricked into thinking there is no actor with that did when there is an actor but sth else wnet wrong

app/facades/trust_registry.py
```diff
@@ -50,3 +145,19 @@ async def get_did_for_actor(actor_id: str) -> List[str]:
...
+    schema_res = requests.post(
+        TRUST_REGISTRY_URL + "/registry/schemas", json={"schema_id": schema_id}
+    )
+
+    if schema_res.status_code != 200:
+        raise Exception(f"Error registering schema {schema_id}: {schema_res.text}")
+
+
+async def register_actor(actor: Actor) -> None:
+    actor_res = requests.post(TRUST_REGISTRY_URL + "/registry/actors", json=actor)
```
morrieinmaas, line 160: Does
```python
json=actor
```
possibly  have to be 
```python
json=actor.dict()
```
or 
```python
json=json.dumps(actor.dict())
```

app/tests/facades/test_trust_registry.py
```diff
@@ -22,7 +130,7 @@ async def test_actor_has_schema():
 
     with patch("requests.get") as mock_request:
         mock_request.return_value.status_code = 200
-        mock_request.return_value.json.return_value = {"schemas": "schema_id"}
+        mock_request.return_value.json.return_value = {"schemas": ["schema_id"]}
```
morrieinmaas, line 133: Nice one!

app/tests/facades/test_trust_registry.py
```diff
@@ -31,31 +139,59 @@ async def test_actor_has_schema():
...
-        mock_request.return_value.json.return_value = {"roles": "issuer"}
+        mock_request.return_value.json.return_value = {"roles": ["issuer"]}
 
         with pytest.raises(HTTPException):
             await trf.actor_has_role("yoma", "issuer")
 
     with patch("requests.get") as mock_request:
         mock_request.return_value.status_code = 200
-        mock_request.return_value.json.return_value = {"roles": "issuer"}
+        mock_request.return_value.json.return_value = {"roles": ["issuer"]}
```
morrieinmaas, line 162: 👏

app/utils.py
```diff
@@ -36,3 +38,12 @@ def construct_indy_proof_request(
         },
     }
     return indy_proof_request
+
+
+def ed25519_verkey_to_did_key(key: str) -> str:
+    """Convert a naked ed25519 verkey to W3C did:key format."""
```
morrieinmaas, line 44: Maybe we can add some mechanism to check whether ed25519 verkey is of valid format so if one passes in a junk key by accident or purpose this fails in a desired way

trustregistry/crud.py
```diff
@@ -2,14 +2,17 @@
...
-from trustregistry import models
-from trustregistry import schemas
+from trustregistry import models, schemas
 
 
 def get_actors(db: Session, skip: int = 0, limit: int = 1000) -> List[models.Actor]:
     return db.query(models.Actor).offset(skip).limit(limit).all()
 
 
+def get_actor_by_did(db: Session, actor_did: str):
```
morrieinmaas, line 12: nice

trustregistry/list_type.py
```diff
@@ -0,0 +1,15 @@
+from sqlalchemy import TypeDecorator
+from sqlalchemy.sql.sqltypes import String
+
+
+class StringList(TypeDecorator):
```
morrieinmaas, line 5: also nice

morrieinmaas (changes requested):
Great job - and nice trick with the custom Type implmentation StringList. 

Left a few remarks you can have a look at. Overall looking great though.

app/utils.py
TimoGlastra: Yes that would be a good one. I'm going to open a separate ticket for it

app/facades/trust_registry.py
TimoGlastra: `Actor` is a typed dict, not a pydantic model. So it is already a dict. I have tested it and it works fine.

morrieinmaas (approved):
nice

Issue #149, opened 2021-11-04
API hangs if no role is provided, should throw error


TimoGlastra:

morrieinmaas:
closed by #162

PR #162, opened 2021-11-08
Fix hanging swagger by:

morrieinmaas:
* Specifying str type for role and
* checking either auth token/key is there

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -30,17 +30,17 @@ services:
       - yoma-ga
       - host
     tty: true
-  ngrok-yoma-ga:
-    image: wernight/ngrok
-    env_file:
-      - environments/yoma-ga/aca-py-agent.default.env
-    command: ngrok http yoma-ga-agent:3020 --log stdout
-    networks:
-      - yoma-ga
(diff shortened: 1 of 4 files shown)


morrieinmaas:
closes #149

docker-compose.yaml
```diff
@@ -30,17 +30,17 @@ services:
...
+  #   image: wernight/ngrok
+  #   env_file:
+  #     - environments/yoma-ga/aca-py-agent.default.env
+  #   command: ngrok http yoma-ga-agent:3020 --log stdout
+  #   networks:
+  #     - yoma-ga
+  #     - host
+  #   depends_on:
+  #     - ledger-nodes
+  #     - ledger-browser
```
TimoGlastra, line 43: Is this needed?

TimoGlastra (approved):

docker-compose.yaml
morrieinmaas: no but we're not using this anyway. Thought we could free up some resources this way on the deploy machine. SHouldn't have snuck it in here I guess.

Issue #154, opened 2021-11-08
ACA-Py configuration for auto-acceptance


TimoGlastra:
To make the process of interacting with the fast api as smooth as possible we should enable some auto accept options, while leaving some other auto configurations turned of for security reasons.

## Checklist

- [ ] Connections
	- `auto-accept-invites: true`
	- `auto-accept-requests: true`
	- `auto-ping-connection: true`
- [ ] Credential & Proofs
	- `preserve-exchange-records: false`
- [ ] Credentials
	- `auto-store-credential: true`
		- Not sure if this is safe because indy doesn't verify the contents, but we can always add methods to get a new credential with correct values
	- `auto-respond-credential-proposal: false`
	- `auto-respond-credential-offer: false`
	- `auto-respond-credential-request: false`
- [ ] Proofs
	- `auto-respond-presentation-proposal: false`
	- `auto-respond-presentation-request: false`
	- `auto-verify-presentation: true`

mrcbond88:
@morrieinmaas Please confirm all is done so we can move this bad boy across

Issue #140, opened 2021-10-25
Inconsisten naming of variables


morrieinmaas:
Following naming in acapy itself tehre are some naming inconsistencies. For example one find `conn_id` and `connection_id` depending on the context. 

There are reasons to strictly follow the same variable names as aca-py. For instance when using types or dicts where such variables appear following aca-py's convention directly let's you use spread operator (**some_type.dict()) without having to to assign each variable manually to the correct name in aca-py.

However, there is also reason to unify namings to be comprehensive, less confusing and consistent e.g either using `conn_id` or `connection_id` consitently (the latter is preferred, less cryptic). 

- [ ] Determine variables that are of same type with different naming
- [ ] Determine which ones, if any, to homogenise and collapse into same variable naming

Issue #129, opened 2021-10-18
how to handle testing when auto accept is enabled in acapy?


TimoGlastra:
The issuer tests still contain a leftover todo related to auto accept being enabled in python. Is auto accept just enabled for development? In that case, maybe we should try to remove the unneeded auto accept configs in ACA-Py to be able to better test.

```
# TODO check for the correct response when state is credential_received
# We can't complete this with auto accept enabled
```

@morrieinmaas any thoughts on this?

## Task List

- [x] Connections (#165)
- [ ] Credentials
- [ ] Proofs -- should be done as part of present proof

morrieinmaas:
@TimoGlastra There were discussions along those lines (no pun intended) already. IMHO this is a bit tricky because depending on whether or not auto-accept si enabled the flow changes. We left it as auot-accept for time being for simplicity reasons. I guess we could create unit tests to mock out either case. for e2e/integration tests we could set a flag and test which case applies, or provide another container with different setting, or restart the container. All of the latter ones don't sound appealing to me. 
I guess either way a step forward is to set auto_accept to false as I doubt True would be a likely setting in production. Then write tests around that.

PR #165, opened 2021-11-09
refactor connections, resolve auto connect

TimoGlastra:
This simplifies the connection api and always enables auto-accept, so should touch some topics in https://github.com/didx-xyz/aries-cloudapi-python/issues/154 and https://github.com/didx-xyz/aries-cloudapi-python/issues/129

Also renames all instances of `invite` to `invitations` as part of https://github.com/didx-xyz/aries-cloudapi-python/issues/140


app/generic/connections/models.py
```diff
@@ -0,0 +1,43 @@
...
+    their_did: Optional[str] = None
+    their_label: Optional[str] = None
+    their_public_did: Optional[str] = None
+    updated_at: Optional[str] = None
+    error_msg: Optional[str] = None
+    invitation_key: Optional[str] = None
+    invitation_msg_id: Optional[str] = None
+
+
+def conn_record_to_connection(connection_record: ConnRecord):
```
morrieinmaas, line 26: Not sure that'd improve teh code here but you could also do sth like (since you're only changing a single key - rfc23_state):

```python
connection_record = connection_record.dict()
connection_record['state'] = connection_record.pop('rfc23_state')  # changes key, keeps value
return Connection(**connection_record)
```

app/generic/connections/connections.py
```diff
@@ -0,0 +1,249 @@
...
+        alias=body.alias,
+        attachments=body.attachments,
+        handshake_protocols=handshake_protocols if body.create_connection else None,
+        use_public_did=body.use_public_did,
+    )
+
+    invitation = await aries_controller.out_of_band.create_invitation(
+        multi_use=body.multi_use,
+        body=oob_body,
+        # FIXME: https://github.com/didx-xyz/aries-cloudcontroller-python/issues/67
```
morrieinmaas, line 94: ah alright I see - was wondering *weird boolean string*

morrieinmaas (changes requested):
Have a look at [codacy](https://app.codacy.com/gh/didx-xyz/aries-cloudapi-python/pullRequest?prid=8384043) and the one relavant comment. Looking very good otherwise - nice work.

app/generic/connections/models.py
TimoGlastra: Hmm, I'm in favour of creating an explicit mapping. Makes it very clear imo how we transform from the ACA-Py model to our own model.

app/generic/connections/connections.py
TimoGlastra: Yep need to fix that. But would like to do a single pass at updating the cloudcontroller and fixing all outstanding bugs

TimoGlastra:
Will look at codacy, forgot that!

app/generic/connections/models.py
morrieinmaas: totally fine by me. was just a suggestion.

morrieinmaas (approved):

Issue #167, opened 2021-11-09
Extract all constants and environments variables into separate file


TimoGlastra:
This makes it a lot easier to get an overview of which environment files are available and being used

PR #171, opened 2021-11-15
feat: extract env vars

morrieinmaas:
Refactor out constants and env vars.
    
This also includes enabling changing the routes/enpoints via the env file instead of manually having to do so in the code.

### DIFF
--- a/environments/yoma-ga/fastapi.default.env
+++ b/environments/yoma-ga/fastapi.default.env
@@ -16,3 +16,30 @@ LEDGER_NETWORK_URL=http://ledger-browser:8000/register
 LEDGER_TYPE=von
 
 PYTHONPATH=/
+
+# Credential definition
+CREDENTIAL_DEFINITIONS_ROUTE=/admin/governance/credential-definitions
+CREDENTIAL_DEFINITIONS_TAGS="admin: credential definitions"
+# Schemas
+SCHEMAS_ROUTE=/admin/governance/schemas
+SCHEMAS_TAGS="admin: schemas"
+# Wallet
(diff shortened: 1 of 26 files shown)


morrieinmaas:
closes #167

TimoGlastra (commented):
Great for picking up this issue. I'm not sure however if we should extract the endpoint and tags into environment variables.  Do you have a specific use case that would require changing the tags/endpoints dynamically? I think it makes it harder to understand the code as you need to follow the environment variables to know for sure what the endpoint should be

Interested to hear your opinion on this

morrieinmaas:
hmm fair point. I'm happy to not do it that way. Reason why I did that is that we have changed the namings and routes 2 already in the past. I thought this'd make it easier (also for non-YOMA) to make the routing and tagging the way they like. But again, fine by me to not do it that way.

environments/yoma-ga/fastapi.default.env
```diff
@@ -15,4 +15,50 @@ ACAPY_TAILS_SERVER_BASE_URL=http://tails-server:6543
 LEDGER_NETWORK_URL=http://ledger-browser:8000/register
 LEDGER_TYPE=von
 
-PYTHONPATH=/
\ No newline at end of file
+# TEST RELATED
+TEST_LEDGER_URL=http://localhost:9000/register"
+CONFTEST_ADMIN_URL = os.getenv("CONFTEST_ADMIN_URL", "http://localhost")
```
TimoGlastra, line 20:
```suggestion
CONFTEST_ADMIN_URL = "http://localhost"
```

morrieinmaas:
Sure. sounds good

TimoGlastra (approved):

PR #172, opened 2021-11-16
Feat/103/request proof

### LINKED ISSUE #103: Request proof

morrieinmaas:
closes #103


morrieinmaas:
@TimoGlastra it's taken a while but eventually... Please, have a look. Looking forward to some feedback. 

Also, not entirely sure whether to do it in this ticket, but I reckon it'd make sense to integrate the trust registry with the endpoints also. Let me know your thoughts on it. Happy to add it in this PR or to open another issue for it.

app/generic/proof/facades/acapy_proof.py
```diff
@@ -0,0 +1,88 @@
...
+        controller: AcaPyClient
+            The aries_cloudcontroller object
+        proof_request: V10PresentationCreateRequestRequest
+            The proof request object
+        """
+
+    @classmethod
+    @abstractmethod
+    async def accept_proof_request(
+        cls, controller: AcaPyClient, pres_ex_id: str, body: Optional[IndyPresSpec]
```
TimoGlastra, line 58: To prevent the issue of using inconsistent abbreviations, maybe we can use the full form `presentation_exchange_id`?

app/generic/proof/facades/acapy_proof.py
```diff
@@ -0,0 +1,88 @@
+from abc import ABC, abstractmethod
+from typing import Optional
+
+from aries_cloudcontroller import AcaPyClient, IndyProofRequest
+from aries_cloudcontroller.model.indy_pres_spec import IndyPresSpec
+
+from app.generic.proof.models import PresentationExchange
+
+
+class Proof(ABC):
```
TimoGlastra, line 10: I'm thinking what the best naming would be for proof. For issue credential we called it `issuer` (which is a role), while `proof` is more a statement. 

We should come up with something consistent for both

- issuer
- prover

or 

- credentials
- proofs

or

- issuer
- verifier

what do you think?

app/generic/proof/facades/acapy_proof_v1.py
```diff
@@ -0,0 +1,180 @@
...
+                presentation_exchange = (
+                    await controller.present_proof_v1_0.send_request(
+                        pres_ex_id=pres_ex_id, body=proof_request
+                    )
+                )
+            elif isinstance(proof_request, V10PresentationProposalRequest):
+                presentation_exchange = (
+                    await controller.present_proof_v1_0.send_proposal(
+                        body=proof_request
+                    )
```
TimoGlastra, line 76: A proposal is different from a request, so it should probably get it's own method. For issuer I just ignored the proposal for now

app/generic/proof/facades/acapy_proof_v1.py
```diff
@@ -0,0 +1,180 @@
...
+                    await controller.present_proof_v1_0.send_request_free(
+                        body=V10PresentationSendRequestRequest(**proof_request.dict())
+                    )
+                )
+            elif isinstance(proof_request, AdminAPIMessageTracing) and pres_ex_id:
+                presentation_exchange = (
+                    await controller.present_proof_v1_0.send_request(
+                        pres_ex_id=pres_ex_id, body=proof_request
+                    )
+                )
```
TimoGlastra, line 71: As I said in the other comment, I would ignore tracing for now. This way there is only one case to handle (the first one)

app/generic/proof/facades/acapy_proof_v1.py
```diff
@@ -0,0 +1,180 @@
...
+            raise e from e
+
+    @classmethod
+    async def accept_proof_request(
+        cls,
+        controller: AcaPyClient,
+        pres_ex_id: str,
+        body: IndyPresSpec,
+    ) -> PresentationExchange:
+        pres_ex_id = cls.__pres_id_no_version(pres_ex_id=pres_ex_id)
```
TimoGlastra, line 91: In the issuer implement the v1 prefix is removed in the issuer endpoints, but I  think how you handled it here  (in the acapy facades) makes more sense (the v1 implementation knows it should remove the `v1-` prefix). 

We should update the issuer implementation to be consistent with your method

app/generic/proof/facades/acapy_proof_v1.py
```diff
@@ -0,0 +1,180 @@
...
+            pres_ex_id=pres_ex_id
+        )
+        if not isinstance(proof_request, V10PresentationExchange) or not isinstance(
+            delete_proof_request_res, (Dict, NoneType)
+        ):
+            raise HTTPException(status_code=500, detail="Failed to delete record")
+
+    @classmethod
+    def __pres_id_no_version(cls, pres_ex_id: str) -> str:
+        if pres_ex_id.startswith("v1-") or pres_ex_id.startswith("v2-"):
```
TimoGlastra, line 131: I'm not sure if we should take v2 into account here, we're in the v1 facade!

app/generic/proof/facades/acapy_proof_v1.py
```diff
@@ -0,0 +1,180 @@
...
+            return pres_ex_id[3:]
+        else:
+            return pres_ex_id
```
TimoGlastra, line 134: I think such convenience methods (support ids with and without `v1-` prefix) may make the API easier in the start, but I think it is good to have a **right** and a **wrong** way to call an API. So I would say we only allow one of the two

app/generic/proof/facades/acapy_proof_v1.py
```diff
@@ -0,0 +1,180 @@
...
+
+    @classmethod
+    def __record_to_model(cls, record: V10PresentationExchange) -> PresentationExchange:
+        # When createing rewuest instead of sending
```
TimoGlastra, line 138:
```suggestion
        # When creating request instead of sending
```

app/generic/proof/facades/acapy_proof_v1.py
```diff
@@ -0,0 +1,180 @@
...
+        if not record.presentation:
+            record.presentation = record.presentation_request_dict.dict()
```
TimoGlastra, line 140: The `presentation` is different from the `request`, so I don't think we should be mixing this

app/generic/proof/facades/acapy_proof_v1.py
```diff
@@ -0,0 +1,180 @@
...
+            presentation=record.presentation,
+            presentation_exchange_id="v1-" + str(record.presentation_exchange_id),
+            role=record.role,
+            state=cls.__v1_state_to_rfc_state(record.state),
+            updated_at=record.updated_at,
+            verified=cls.__string_to_bool(record.verified),
+        )
+
+    @classmethod
+    def __string_to_bool(cls, verified: Optional[str]) -> Optional[bool]:
```
TimoGlastra, line 156: 😁

app/generic/proof/facades/acapy_proof_v2.py
```diff
@@ -0,0 +1,159 @@
...
+    V20PresSpecByFormatRequest,
+)
+from fastapi.exceptions import HTTPException
+from pydantic.typing import NoneType
+
+from app.generic.proof.facades.acapy_proof import Proof
+from app.generic.proof.models import PresentationExchange
+
+
+class ProofsV2(Proof):
```
TimoGlastra, line 21: Comments are mostly the same as on the v1 version

app/generic/proof/facades/acapy_proof_v2.py
```diff
@@ -0,0 +1,159 @@
...
+        if pres_ex_id.startswith("v1-") or pres_ex_id.startswith("v2-"):
+            return pres_ex_id[3:]
+        else:
+            return pres_ex_id
+
+    @classmethod
+    def __record_to_model(cls, record: V20PresExRecord) -> PresentationExchange:
+
+        return PresentationExchange(
+            auto_present=record.auto_present,
```
TimoGlastra, line 122: I think as we're not allowing to set this through the api we should not include `auto-present`

app/generic/proof/facades/acapy_proof_v2.py
```diff
@@ -0,0 +1,159 @@
...
+    def __pres_id_no_version(cls, pres_ex_id: str) -> str:
+        if pres_ex_id.startswith("v1-") or pres_ex_id.startswith("v2-"):
+            return pres_ex_id[3:]
+        else:
+            return pres_ex_id
+
+    @classmethod
+    def __record_to_model(cls, record: V20PresExRecord) -> PresentationExchange:
+
+        return PresentationExchange(
```
TimoGlastra, line 121: In the issuer api I added an `protocol_version`  attribute and `IssueCredentialProtocolVersion` enum to indicate the version of the protocol used for the exchange

app/generic/proof/facades/acapy_proof_v2.py
```diff
@@ -0,0 +1,159 @@
...
+            auto_present=record.auto_present,
+            connection_id=record.connection_id,
+            created_at=record.created_at,
+            initiator=record.initiator,
```
TimoGlastra, line 125: We don't have initiator on the issuer record, so If we want to add it here (you can extract it from the state, but I can definitely see the value of adding it), we should also add it to the issuer api

app/generic/proof/facades/acapy_proof_v2.py
```diff
@@ -0,0 +1,159 @@
...
+        )
+
+    @classmethod
+    def __string_to_bool(cls, verified: Optional[str]) -> Optional[bool]:
+        if verified == "true":
+            return True
+        elif verified == "false":
+            return False
+        else:
+            return None
```
TimoGlastra, line 141: Extract to separate util function that both versions can use?

app/generic/proof/models.py
```diff
@@ -0,0 +1,40 @@
...
+    V20PresExRecordByFormat,
+    V20Pres,
+    IndyProof,
+)
+from pydantic import BaseModel
+
+
+class ProofRequestProtocolVersion(Enum):
+    v10 = "v1"
+    v20 = "v2"
```
TimoGlastra, line 14: Ah I see you added it here, but it's not in the record?

app/generic/proof/proof.py
```diff
@@ -0,0 +1,217 @@
...
+from app.generic.proof.facades.acapy_proof_v1 import ProofsV1
+from app.generic.proof.facades.acapy_proof_v2 import ProofsV2
+from app.generic.proof.models import PresentationExchange
+
+logger = logging.getLogger(__name__)
+
+
+router = APIRouter(prefix="/generic/proof", tags=["proof"])
+
+ProtocolVersion = Literal["1", "2"]
```
TimoGlastra, line 26: I think we can reuse the `ProofRequestProtocolVersion` here

app/generic/proof/proof.py
```diff
@@ -0,0 +1,217 @@
...
+    v10 = ProofsV1
+    v20 = ProofsV2
+
+
+@router.post("/send-request")
+async def send_proof_request(
+    connection_id: str,
+    proof_request: IndyProofRequest,
+    protocol_version: Optional[ProtocolVersion] = "1",
+    aries_controller: AcaPyClient = Depends(agent_selector),
```
TimoGlastra, line 39: Please see the api of the issuer endpoints. I integrated the connection id and protocol version into the body of the request.

This method is also fine, but I think above all we should be consistent in how the API looks. 

```
POST /generic/issuer/credentials
```

vs

```
POST /generic/proof/send-request?protocol_version=2&connection_id=xxxx
```

app/generic/proof/proof.py
```diff
@@ -0,0 +1,217 @@
...
+            return presentation_exchange_rec
+        else:
+            presentation_exchange_rec = await ProofsFacade.v10.value.send_proof_request(
+                controller=aries_controller,
+                proof_request=V10PresentationSendRequestRequest(
+                    connection_id=connection_id,
+                    proof_request=IndyProofRequest(**proof_request.dict()),
+                ),
+                free=True,  # alway set this to TRUE because we only support this for now.
+            )
```
TimoGlastra, line 79: One of the ideas of the abstract interface is that we don't need the if else statement but abstract everything away that's different between the versions. See the issuer implementation for an example.

E.g. the different between v1 and v2 can be handled by the facade implementation, while the common structure (`IndyProofRequest`) can be handled by this layer

app/generic/proof/proof.py
```diff
@@ -0,0 +1,217 @@
...
+        if protocol_version == "2":
+            presentation_exchange_rec = await ProofsFacade.v20.value.send_proof_request(
+                controller=aries_controller,
+                proof_request=V20PresSendRequestRequest(
+                    connection_id=connection_id,
+                    presentation_request=V20PresRequestByFormat(
+                        dif=None, indy=IndyProofRequest(**proof_request.dict())
+                    ),
+                ),
+                free=True,  # alway set this to TRUE because we only support this for now.
```
TimoGlastra, line 68: I'd just remove the free and always do that (see my other comments)

app/generic/proof/proof.py
```diff
@@ -0,0 +1,217 @@
...
+class ProofsFacade(Enum):
+    v10 = ProofsV1
+    v20 = ProofsV2
+
+
+@router.post("/send-request")
+async def send_proof_request(
+    connection_id: str,
+    proof_request: IndyProofRequest,
+    protocol_version: Optional[ProtocolVersion] = "1",
```
TimoGlastra, line 38: I think there can be value in making the protocol_version required. this way it is more explicit to the user which version they are using.

app/generic/proof/proof.py
```diff
@@ -0,0 +1,217 @@
...
+        return presentation_exchange_rec
+    except Exception as e:
+        logger.error(f"Failed to create presentation record: \n{e!r}")
+        raise e from e
+
+
+@router.post("/create-request")
+async def create_proof_request(
+    proof_request: IndyProofRequest,
+    comment: Optional[str] = None,
```
TimoGlastra, line 89: I'd combine as much parameters as possible in the body and nest them to be consistent with the issuer api

```
{
   "comment": "xxx",
   "proof_request": {...}
}

File: app/generic/proof/proof.py [hunk 23]
```diff
@@ -0,0 +1,217 @@
...
+                ),
+                free=True,  # alway set this to TRUE because we only support this for now.
+            )
+            return presentation_exchange_rec
+        else:
+            presentation_exchange_rec = await ProofsFacade.v10.value.send_proof_request(
+                controller=aries_controller,
+                proof_request=V10PresentationSendRequestRequest(
+                    connection_id=connection_id,
+                    proof_request=IndyProofRequest(**proof_request.dict()),
```
TimoGlastra, line 76:
```suggestion
                    proof_request=IndyProofRequest(proof_request),
```

app/tests/e2e/test_proof.py
```diff
@@ -0,0 +1,287 @@
...
+                "support_revocation": False,
+                "schema_id": schema_def["schema_id"],
+                "tag": get_random_string(5),
+            }
+        ),
+    )
+    result = response.json()
+
+    print(f"created definition {str(result)}")
+    return result["credential_definition_id"]
```
TimoGlastra, line 68: This is copied exactly from the issuer tests it seems. I think we should extract them into a seperate file to avoid duplication

app/tests/e2e/test_proof.py
```diff
@@ -0,0 +1,287 @@
...
+
+
+@pytest.fixture(scope="module")
+async def credential_exchange_id(
+    async_client_bob_module_scope: AsyncClient,
```
TimoGlastra, line 73: Same also duplicated

app/tests/e2e/test_proof.py
```diff
@@ -0,0 +1,287 @@
...
+
+@pytest.mark.asyncio
+async def test_send_proof_request(
+    alice_connection_id: str,
+    async_client_alice_module_scope: AsyncClient,
+):
+    # V1
+    response = await async_client_alice_module_scope.post(
+        BASE_PATH + f"/send-request?connection_id={alice_connection_id}",
+        data=json.dumps(proof_dict),
```
TimoGlastra, line 145:
```suggestion
        json=proof_dict,
```

app/tests/e2e/test_proof.py
```diff
@@ -0,0 +1,287 @@
...
+        + f"/send-request?connection_id={alice_connection_id}&protocol_version=2",
+        data=json.dumps(proof_dict),
+    )
+
+    result = response.json()
+    assert "auto_present" is not None
+    assert "presentation" in result.keys()
+    assert "created_at" in result.keys()
+    assert "presentation_exchange_id" in result.keys()
+    assert "v2-" in result["presentation_exchange_id"]
```
TimoGlastra, line 168: In the issuer implementation I used `credential_id` insted of `credential_exchange_id`. Maybe we can here also use `proof_id` (or whather we are going to  call it instead of proof) instead of `presentation_exchange_id`. This feels more in line with the rest of the api

app/trust_registry_facade.py
```diff
@@ -0,0 +1,52 @@
+import logging
+import os
+from typing import List, Literal
+
+import requests
+from fastapi.exceptions import HTTPException
+
+TRUST_REGISTRY_URL = os.getenv("TRUST_REGISTRY_URL", "http://localhost:8001/")
+
+logger = logging.getLogger(__name__)
```
TimoGlastra, line 10: This is now hosted in `app/facades/trust_registry.py`

junit.xml
```diff
@@ -0,0 +1,112 @@
+<?xml version="1.0" encoding="utf-8"?><testsuites><testsuite name="pytest" errors="0" failures="1" skipped="0" tests="4" time="25.857" timestamp="2021-11-22T17:42:32.382209" hostname="MBP-van-Moritz"><testcase classname="tests.e2e.test_proof" name="test_send_proof_request" time="25.262" /><testcase classname="tests.e2e.test_proof" name="test_create_proof_request" time="0.041" /><testcase classname="tests.e2e.test_proof" name="test_accept_proof_request" time="0.060"><failure message="TypeError: must be real number, not ClientResponseError">alice_connection_id = '5a30ab64-7e5e-45f4-b132-ed857ed71e5d', async_client_alice_module_scope = &lt;httpx.AsyncClient object at 0x10f51c070&gt;
```
TimoGlastra, line 1: I think I added this to `.gitignore`? Don't think it belongs in the PR

TimoGlastra (changes requested):
Nice work @morrieinmaas!

I left a bunch of comments, but most are related to the following two topics:
- Abstraction of v1 and v2 specific details on higher levels
- Consistency with the issuer api

morrieinmaas:
Nice. Thnaks for all the feedback. I'll get it all addressed.

app/generic/proof/facades/acapy_proof.py
TimoGlastra: Forget that, in a later comment I suggested using `proof_id` instead of using the acapy naming (to be consistent with issuer api)

app/generic/proof/facades/acapy_proof_v1.py
morrieinmaas: cool. I'll do that in this PR then.

app/generic/proof/facades/acapy_proof_v1.py
morrieinmaas: agreed

app/generic/proof/facades/acapy_proof_v2.py
morrieinmaas: Can also do that in this PR. OR a different PR, @TimoGlastra ?

app/generic/proof/facades/acapy_proof_v2.py
morrieinmaas: yeah, fair point

app/generic/proof/proof.py
morrieinmaas: That was my thinking also. One the one hand, it does require a more conscious choice. On the other hand, though, might save some frustration to provide a default one. Not quite sure which way is better, @TimoGlastra

junit.xml
morrieinmaas: hmm yeah I thought so, too, tbh.

app/generic/proof/facades/acapy_proof.py
morrieinmaas: hmm, that's a very good point actually. lt me have a think about it.

app/generic/proof/facades/acapy_proof_v2.py
morrieinmaas: i'll just remove it for now

app/generic/proof/facades/acapy_proof.py
morrieinmaas: went for issuer/verifier combo

app/generic/proof/facades/acapy_proof_v1.py
morrieinmaas: Will do in fidderent PR actually. There's too much in this one otherwise, imho

app/admin/governance/multitenant_wallet/wallet_admin.py
```diff
@@ -10,6 +10,7 @@
 from fastapi import APIRouter, Depends, HTTPException
 from pydantic import BaseModel
 
+# from app.config import get_settings
```
TimoGlastra, line 13: ?

app/generic/verifier/facades/acapy_verifier_utils.py
```diff
@@ -0,0 +1,73 @@
...
+    translation_dict = {
+        "proposal_sent": "proposal-sent",
+        "proposal_received": "proposal-received",
+        "request_sent": "request-sent",
+        "request_received": "request-received",
+        "presentation_sent": "presentation-sent",
+        "presentation_received": "presentation-received",
+        "done": "done",
+        "abandoned": "abandoned",
+    }
```
TimoGlastra, line 68: Is the start state the same for both v1 and v2?

app/generic/verifier/models.py
```diff
@@ -0,0 +1,66 @@
...
+class PresentationExchange(BaseModel):
+    connection_id: Optional[str] = None
+    created_at: Optional[str] = None
+    proof_id: str
+    presentation: Optional[
+        Union[IndyProof, V20Pres, V20PresExRecordByFormat, Dict]
+    ] = None
+    protocol_version: ProofRequestProtocolVersion
+    role: Literal["prover", "verifier"]
+    state: Optional[
```
TimoGlastra, line 53: State is always defined

```suggestion
    state: 
```

app/generic/verifier/models.py
```diff
@@ -0,0 +1,66 @@
...
+    problem_report: Optional[str] = None
+
+
+class PresentationExchange(BaseModel):
+    connection_id: Optional[str] = None
+    created_at: Optional[str] = None
+    proof_id: str
+    presentation: Optional[
+        Union[IndyProof, V20Pres, V20PresExRecordByFormat, Dict]
+    ] = None
```
TimoGlastra, line 50: I still think it would be good to unify this. We can schedule a call to look into what that means if you want?

app/generic/verifier/models.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+class RejectProofRequest(ProofRequestBase):
+    proof_id: Optional[str] = None
+    problem_report: Optional[str] = None
+
+
+class PresentationExchange(BaseModel):
+    connection_id: Optional[str] = None
+    created_at: Optional[str] = None
```
TimoGlastra, line 46: Created at is not optional

app/generic/verifier/verifier.py
```diff
@@ -0,0 +1,154 @@
...
+    PresentationExchange,
+    ProofRequestProtocolVersion,
+    RejectProofRequest,
+    SendProofRequest,
+)
+
+logger = logging.getLogger(__name__)
+
+
+router = APIRouter(prefix="/generic/proof", tags=["proof"])
```
TimoGlastra, line 24: This isn't updated to verifier yet

app/generic/verifier/verifier.py
```diff
@@ -0,0 +1,154 @@
...
+    proof_request: SendProofRequest
+        The proof request
+
+    Returns:
+    --------
+    presentation_exchange: PresentationExchange
+        The presentation exchange record
+    """
+    try:
+        prover = __get_prover_by_version(proof_request.protocol_version)
```
TimoGlastra, line 64:
```suggestion
        prover = __get_verifier_by_version(proof_request.protocol_version)
```

app/generic/verifier/verifier.py
```diff
@@ -0,0 +1,154 @@
...
+)
+
+logger = logging.getLogger(__name__)
+
+
+router = APIRouter(prefix="/generic/proof", tags=["proof"])
+
+ProtocolVersion = Literal[
+    ProofRequestProtocolVersion.v10.value, ProofRequestProtocolVersion.v20.value
+]
```
TimoGlastra, line 28: Is this used? I think we can just use the protocol version values right?

app/generic/verifier/verifier.py
```diff
@@ -0,0 +1,154 @@
...
+    proof_request: AcceptProofRequest
+        The proof request
+
+    Returns:
+    --------
+    presentation_exchange: PresentationExchange
+        The presentation exchange record
+    """
+    try:
+        prover = __get_prover_by_version(proof_request.protocol_version)
```
TimoGlastra, line 120: If the proof request already exists we can determine the protocol version based on the id (`v1-` vs `v2-` prefix)

dockerfiles/fastapi/Dockerfile
```diff
@@ -4,7 +4,7 @@ COPY app /app
 
 WORKDIR /app
 
-RUN cd /app && pip install -r requirements.txt
+RUN cd /app && pip install -r requirements.txt && export $(grep -v '^#' /app/.env | xargs)
```
TimoGlastra, line 7: What does this do?

morrieinmaas:
Just trying to figure out what is going wrong in the build pipeline. The tests that are failing there work fine for me

```bash
pytest -s -vvv app/tests/e2e/test_verifier.py
===================================================================================== test session starts ======================================================================================
platform darwin -- Python 3.8.10, pytest-6.2.4, py-1.10.0, pluggy-1.0.0.dev0 -- /Users/moriarty/.pyenv/versions/3.8.10/bin/python3.8
rootdir: /Users/moriarty/Projects/aries-cloudapi-python/app, configfile: setup.cfg
plugins: cov-2.12.1, bdd-5.0.0, asyncio-0.15.1, anyio-3.3.4
collected 4 items                                                                                                                                                                              

app/tests/e2e/test_verifier.py::test_send_proof_request PASSED
app/tests/e2e/test_verifier.py::test_create_proof_request PASSED
app/tests/e2e/test_verifier.py::test_accept_proof_request PASSED
app/tests/e2e/test_verifier.py::test_reject_proof_request PASSED

--------------------------------------------------------- generated xml file: /Users/moriarty/Projects/aries-cloudapi-python/junit.xml ---------------------------------------------------------
```

TimoGlastra (commented):
Really great improvements. Some small nitpicks left.

I think the main thing is that we should unify the type of `presentation`

app/admin/governance/multitenant_wallet/wallet_admin.py
morrieinmaas: ?... not sure what happened there. sorry

app/generic/verifier/models.py
morrieinmaas: yes let's do that

app/generic/verifier/verifier.py
morrieinmaas: hmm. yeah we could also do that.

dockerfiles/fastapi/Dockerfile
morrieinmaas: don't know where that is coming from either

morrieinmaas:
Thanks for all the help and review comments. The feedback helped improve the code immnsely.

dockerfiles/fastapi/Dockerfile
morrieinmaas: Alright I remember now where this is from, but not sure how that ended up in the commit. That was from the other ticket about extracting all env vars, but this bit never got merged. I'll remove it

app/generic/verifier/facades/acapy_verifier_utils.py
morrieinmaas: Can quickly discuss this also in the call for unifying the presentation. Not 100% sure I know what you mean.

morrieinmaas: This is only used for v1. v2 has the `-` instead of the `_` already.

TimoGlastra (approved):

TimoGlastra:
Awesome! really great work. I think we can merge and do the outstanding tweaks in follow up PRs!

Issue #173, opened 2021-11-17
use single library for http requests


TimoGlastra:
currently `httpx` is used for tests, while `requests` is used for the normal app. I think we should use a single HTTP client library for both

morrieinmaas:
agreed

TimoGlastra:
Do you have a preference between the two?  As I understand it now `requests` is very well established and used everywhere while `httpx` is the new kid on the block with nice new features.

But I haven't worked a lot with both of them, so interested to hear your opinion

morrieinmaas:
Hmm, was going to say I don't but tunrs out I do. I'd go with httpx. It's really just as easy to use as requests and comes with some extra niceties.

michaelwiles:
I believe it was I that originally used httpx for http requests in tests.

The reason I went with httpx is that it supports async and [requests doesn't](https://stackoverflow.com/questions/9110593/asynchronous-requests-with-python-requests).

michaelwiles:
I just remembered... the other reason for httpx is because the examples of how to test a fastapi back end were given in httpx.

PR #178, opened 2021-11-17
feat: use httpx as single http request library

morrieinmaas:
replace requests calls by httpx calls and remove requests from the requirements.

### DIFF
--- a/app/facades/ledger.py
+++ b/app/facades/ledger.py
@@ -2,7 +2,7 @@ import logging
 import os
 from typing import Optional
 
-import requests
+import httpx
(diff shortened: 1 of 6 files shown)


morrieinmaas:
closes #173

TimoGlastra (approved):
Don't we need to add httpx to the app dependencies.txt?

Otherwise LGTM!

morrieinmaas:
argh well spotted! wasn't an issue for me because it was installed in my virtual env anyway. You mean requirements.txt?!

Lemme push that

PR #179, opened 2021-11-17
Remove inital wait/sleep in github actions.

### LINKED ISSUE #174: Find way to remove sleeps from CI pipeline
It adds a lot of time to the time it takes to run tests, and I think it should be possible to use some tricks to check whether we can continue instead of using sleeps. Not sure how yet...

morrieinmaas:
Replace it by checking the logs of one of the aca-py containers for signalling ready.

Best I could come up with for now. That's what works for me on local machine. 

Happy if there are better suggestions. Couldn't find any poll events to look for in particular as suggested in the ticket.

closes #174

### DIFF
--- a/.github/workflows/continuous-build.yml
+++ b/.github/workflows/continuous-build.yml
@@ -30,18 +30,18 @@ jobs:
       - name: Start Test Harness
         run: |
           chmod u+x manage
-          ./manage up-as-daemon
+          ./manage up-daemon-usecache
           i=0
+          echo echo waiting for dockers to startup
           while [ $i -le 2 ]
           do
-            echo echo waiting for dockers to startup
             if (docker logs $(docker ps -f name="yoma-ga-agent" | awk 'FNR == 2 {print $1}') | grep -q "::::::::::::::::::::::::::::::::::::::::::::::
             :: Alice                                    ::
             ::                                          ::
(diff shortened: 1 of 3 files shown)


morrieinmaas:
also I left the sleep in for the retry on purpose in a better-safe-than-sorry kinda way. Can also change that if desired.

morrieinmaas:
Runs through now. Saves roughly 3-6 minutes compared to previous implementation when looking at the entire runtime of the ci/cd pipeline runs under the actions tab

TimoGlastra (approved):
That looks complex 😬

morrieinmaas:
It;s really just the only thing i could think of right away. Not that complex really. Just checks the logs of aca-py container for this "ascii art" bit aca-py logs when ready to go:

```bash
::::::::::::::::::::::::::::::::::::::::::::::
:: Alice                                    ::
::                                          ::
::                                          ::
:: Inbound Transports:                      ::
::                                          ::
::   - http://0.0.0.0:3020                  ::
::                                          ::
:: Outbound Transports:                     ::
::                                          ::
::   - http                                 ::
::   - https                                ::
::                                          ::
:: Public DID Information:                  ::
::                                          ::
::   - DID ....
```

when that happens it's ready to go. That's pretty much it.

PR #180, opened 2021-11-22
refactor: use fastapi authentication

TimoGlastra:
Use fastapi authentication method. This makes it a lot easier to interact with the swagger ui as it adds an authorize button to the top that work for all requests.

image

I updated the `x-auth` parameter to `Authorization` as FastAPI doesn't allow two api key header params. 

As a next step I would like to look at integrating the role into the api key. This way you don't have to specify the role yourself, but rather this is determined by the api key. What do you think about that @morrieinmaas?


morrieinmaas:
@TimoGlastra That's great I think. Really nice improvement.

TimoGlastra:
@morrieinmaas  this is now ready for review.

There are still a lot of codacy issues, but 99% of them are unused/overwritten names because of how pytest fixtures work.

The scope of this PR has expanded quite a bit. The initial plan was to revamp the authentication mechanism a bit, but to do that I had to change lot of the underlying structure, which got quite complicated. That's why I tried to clean up some things, and also split out the conftest into multiple files. There is a seperate persona file now that we use for bob and alice scenarios. In the future we can expand with (this will become handy with the ecosystem task I'm working on). 

Also, all authentication and roles are now combined into a single parameter. The combination of all parameters became quite complex to manage. The structure is now:

```
x-api-key {role}.{token}
```

e.g. for yoma it would be

```
x-api-key yoma.adminApiKey
```

for a member it can be

```
x-api-key member.eyxxxatokenhere
```

I also split out the admin roles for member and ecosystem to make it easier to defines which roles can call an api.

an admin can now call the api using

```
x-api-key member-admin.adminApiKey
```

Let me know what you think

morrieinmaas (commented):
Nice one. This contains some excellent improvemnts and I almost hate to say it, but there is little critical comments to make. 

I did go through the flagged up (by codacy) issues though and removed some of the imports there that actually weren't needed. Ran all tests locally to ensure everything passes still.

TimoGlastra:
Awesome! I left the imports there on purpose to make it more explicit where the methods were coming from, but maybe that's not the way to do it... Just feels like too much magic how fixtures work in pytest....

morrieinmaas:
Ah right I see. Didn't occur to me that was the reason you left th there.
Understand how that can seem like magic otherwise. Happy to put them back
if you want. Just removed them cos it got a bit out of hand/messy with
codacy issues and using pyflakes locally

TimoGlastra:
Yes I agree, it is better to leave them out!

S

PR #184, opened 2021-11-25
Homogenise presentation format to indy (Dict) in PresentationExchange

### LINKED ISSUE #183: unify presentation value type for PresentationExchange v1/v2

morrieinmaas:
Add presentation_request to PresentationExchange

closes #183

### DIFF
--- a/app/generic/verifier/facades/acapy_verifier_utils.py
+++ b/app/generic/verifier/facades/acapy_verifier_utils.py
@@ -3,6 +3,8 @@ from typing import Optional, Union
 from aries_cloudcontroller import V10PresentationExchange, V20PresExRecord
 
 from app.generic.verifier.models import (
+    IndyProof,
+    IndyProofRequest,
     PresentationExchange,
     ProofRequestProtocolVersion,
 )
@@ -23,10 +25,12 @@ def record_to_model(
             connection_id=record.connection_id,
             created_at=record.created_at,
             protocol_version=ProofRequestProtocolVersion.v20.value,
-            presentation=record.by_format.pres["indy"]
+            presentation=IndyProof(**record.by_format.pres["indy"])
             if record.by_format.pres
-            else record.by_format.pres,
-            presentation_request=record.by_format.pres_request["indy"],
+            else None,
+            presentation_request=IndyProofRequest(
+                **record.by_format.pres_request["indy"]
+            ),
             proof_id="v2-" + str(record.pres_ex_id),
             role=record.role,
             state=record.state,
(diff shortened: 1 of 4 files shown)


app/generic/verifier/facades/acapy_verifier_utils.py
```diff
@@ -23,7 +23,10 @@ def record_to_model(
             connection_id=record.connection_id,
             created_at=record.created_at,
             protocol_version=ProofRequestProtocolVersion.v20.value,
-            presentation=record.pres,
+            presentation=record.by_format.pres["indy"]
+            if record.by_format.pres
+            else record.by_format.pres,
```
morrieinmaas, line 28: If this is None (no presentation was provided)

app/generic/verifier/facades/acapy_verifier_utils.py
```diff
@@ -23,7 +23,10 @@ def record_to_model(
...
+            presentation_request=record.by_format.pres_request["indy"],
```
morrieinmaas, line 29: Request was still made

app/generic/verifier/facades/acapy_verifier_utils.py
TimoGlastra, line 28: `record.by_format.pres` will always be `None` right? In that case we can maybe use `None` to make it more explicit?

```suggestion
            else None
```

app/generic/verifier/models.py
```diff
@@ -45,9 +42,8 @@ class PresentationExchange(BaseModel):
     connection_id: Optional[str] = None
     created_at: str
     proof_id: str
-    presentation: Optional[
-        Union[IndyProof, V20Pres, V20PresExRecordByFormat, Dict]
-    ] = None
+    presentation: Optional[Dict] = None
+    presentation_request: Optional[Dict] = None
```
TimoGlastra, line 46: If we use the model types here we would also have typing in the swagger ui:

```suggestion
    presentation: Optional[IndyProof] = None
    presentation_request: Optional[IndyProofRequest] = None
```

TimoGlastra (commented):
Nice! two nitpicks

app/generic/verifier/models.py
morrieinmaas: sure I'll give that a try. I think aca-py returns a dict so I'll have to construct that from the dict. will give it a go

app/generic/verifier/facades/acapy_verifier_utils.py
morrieinmaas: fair point

app/generic/verifier/models.py
TimoGlastra: We do have to transform the dict to a model then in the v2 record_to_model  transformation

PR #193, opened 2021-12-06
feat: add missing verifier endpoints

### LINKED ISSUE #191: Verifier endpoints are missing read / delete operations
The verifier route is missing get by id, get all, delete, etc... (like we have in issuer)

morrieinmaas:

closes #191


app/generic/verifier/facades/acapy_verifier.py
```diff
@@ -87,17 +92,74 @@ async def reject_proof_request(
...
+    async def delete_proof(cls, controller: AcaPyClient, proof_id: str) -> None:
+        """
+        Delete proof request
+
+        Parameters:
+        -----------
+        controller: AcaPyClient
+            The aries_cloudcontroller object
+        pres_ex_id: str
+            The proof record exchange id
```
TimoGlastra, line 121: outdated comment

app/generic/verifier/facades/acapy_verifier.py
```diff
@@ -87,17 +92,74 @@ async def reject_proof_request(
...
+
         Returns:
         --------
         None
             Returns None on successful record deletion.
         """
+
+    @classmethod
+    @abstractmethod
+    async def proofs(
```
TimoGlastra, line 131: To be consistent with issuer (or also change the issuer class `get_records` is not the best name prob.). But consistency is key :)

```suggestion
    async def get_records(
```

app/generic/verifier/facades/acapy_verifier_v1.py
```diff
@@ -45,6 +45,51 @@ async def create_proof_request(
         )
         return utils.record_to_model(presentation_exchange)
 
+    @classmethod
+    async def get_creds(cls, controller: AcaPyClient, pres_ex_id: str):
```
TimoGlastra, line 49: In `acapy_verifier` it is called `get_credentials` and here `get_creds`? 

Maybe call it `get_credentials_for_request` to be as explicit as we can be

app/generic/verifier/facades/acapy_verifier_v1.py
```diff
@@ -45,6 +45,51 @@ async def create_proof_request(
...
+            return await controller.present_proof_v1_0.get_matching_credentials(
+                pres_ex_id=proof_id
+            )
+        except Exception as e:
+            logger.error(f"{e!r}")
+            raise e from e
+
+    @classmethod
+    async def get_proofs(
+        cls, controller: AcaPyClient, pres_ex_id: Optional[str] = None
```
TimoGlastra, line 61: I would split this method up in two separate methods. One for get proofs and one for get proof by id. This is also how we did it for issuer, and also how the ACA-Py API works.

TimoGlastra: Also I think we should be consistent in using `pres_ex_id` vs `proof_id`

app/generic/verifier/verifier.py
```diff
@@ -36,6 +39,103 @@ def __get_verifier_by_version(protocol_version: str) -> Verifier:
         raise ValueError(f"Unknown protocol version {protocol_version}")
 
 
+@router.post("/credentials")
+async def get_credentials(
+    proof_request: ProofRequestGeneric,
```
TimoGlastra, line 44: `ProofRequestGeneric` has `proof_id` as optional parameter, but it is required for this endpoint. Maybe we can make a separate type that marks the `proof_id` as required?

app/generic/verifier/verifier.py
```diff
@@ -36,6 +39,103 @@ def __get_verifier_by_version(protocol_version: str) -> Verifier:
...
+        )
+    except Exception as e:
+        logger.error(
+            f"Failed to get matching credentials: \n {proof_request.proof_id}\n \n{e!r}"
+        )
+        raise e from e
+
+
+@router.post("/proofs")
+async def get_proofs(
```
TimoGlastra, line 75: Same as with the acapy_verifier I would split this up in multiple endpoints (get proof by id and get proofs).

app/generic/verifier/verifier.py
```diff
@@ -36,6 +39,103 @@ def __get_verifier_by_version(protocol_version: str) -> Verifier:
...
+    Returns:
+    --------
+    None
+    """
+    try:
+        # Currently proof_id[0:2] is required for otherwise the protocol version is unknown
+        # Maybe there is a better way to handle this?
+        # Possibly also handling proof-di not starting with v1- or v2-?
+        # Would it make sense to assume it is v1- by default/if none is supplied
+        prover = __get_verifier_by_version(protocol_version=proof_id[0:2])
```
TimoGlastra, line 131: Maybe we can add an additional util method `__get_verifier_by_id` that does this internally?

TimoGlastra (changes requested):
Nice @morrieinmaas. 

Left some comments

app/generic/verifier/facades/acapy_verifier.py
morrieinmaas: 😅

app/generic/verifier/facades/acapy_verifier_v1.py
morrieinmaas: sure, fine by me. Put that into one because I thought that's more convenient, but have no preference here/happy to split it into 2.

app/generic/verifier/verifier.py
morrieinmaas: Let me have a look, can't think of it rn from the top of my head, but there was a reason I did it this way... Happy to change it to your suggestion, though.

app/generic/verifier/facades/acapy_verifier_v1.py
morrieinmaas: > Also I think we should be consistent in using `pres_ex_id` vs `proof_id`

This btw I chose because the PresentationExchange uses proof_id. Thought this would make more sense to the user when getting a PresentationExchange with a proof ID (that contains v1- or v2-) to put that in knowing what is meant. Under the hood it's still pres_ex_id for the background methods. Not sure what's best here. Any preference on how to go about this @TimoGlastra?

app/generic/verifier/verifier.py
morrieinmaas: made it required now

TimoGlastra: Hmm. `ProofRequestGeneric` depends on `ProofRequestBase` which has a  `protocol_version`. I don't think that's necessary for this call right, as we can extract it from the `proof_id`?

app/generic/verifier/verifier.py
```diff
@@ -27,13 +31,129 @@ class VerifierFacade(Enum):
...
+         The proof request object
+
+     Returns:
+     --------
+     presentation_exchange_list: [IndyCredPrecis]
+         The list of Indy presentation credentials
+    """
+    try:
+        prover = __get_verifier_by_version(
+            version_candidate=proof_request.protocol_version
```
TimoGlastra, line 63: We can extract the `protocol_version` from the `proof_id`. As now it is possible to pass `proof_id=v1-xxx` and `protocol_version='v2'`, which can lead to weird inconsistencies

app/generic/verifier/verifier.py
```diff
@@ -27,13 +31,129 @@ class VerifierFacade(Enum):
...
+            controller=aries_controller, pres_ex_id=proof_request.proof_id
+        )
+    except Exception as e:
+        logger.error(
+            f"Failed to get matching credentials: \n {proof_request.proof_id}\n \n{e!r}"
+        )
+        raise e from e
+
+
+@router.post("/get-records")
```
TimoGlastra, line 75: Sorry to be so pedantic again, but I think having the api just `router.get('/proofs')` is more in line with REST principles (https://jsonapi.org) and is also consistent with issuer. Then for `get_proof_record` we do `router.get('/proofs/{proof_id}')`.

If you want to go the `/get-records` and `/get-record` route you can probably convince me, but we should also change the issuer endpoint then.

I think we should put consistency as a first principle in the cloud api, As consistency makes things easy to use.

app/generic/verifier/verifier.py
```diff
@@ -27,13 +31,129 @@ class VerifierFacade(Enum):
...
+async def get_proof_records(
+    proof_request: ProofRequestBase,
+    aries_controller: AcaPyClient = Depends(agent_selector),
+) -> List[PresentationExchange]:
+    """
+    Get a specific proof record
```
TimoGlastra, line 81: outdated comment

TimoGlastra (changes requested):

app/generic/verifier/verifier.py
morrieinmaas: no need to be sorry. makes it better. I'm on it.

app/generic/verifier/verifier.py
morrieinmaas: 🤦‍♂️ - will change

app/generic/verifier/verifier.py
morrieinmaas: the only hting I'm thinking now is that for the get all with get('/proofs') how are we gonna specify the protocol version?

morrieinmaas: i guess we could do sth like `get('/proofs/{version}'` bu tthat would make it inconsistent with what we've done so far. Maybe I'm not seeing the obvious...

morrieinmaas: never mind actually got an idea

TimoGlastra: In the issuer I didn't include a version specifier. But if we want, we could add it through query parameters like filters

morrieinmaas: I've done as you've done it in issuer for now. I guess we could also return a custom type/class with v1: [v1_proofs] and v2: [v2_proofs]

TimoGlastra: Hmm I would try to stay away from that as the protocol version being used is not really important IMO. As an entity using this API I want to request a proof. Whether that is done using v1 or v2 can have the exact same outcome. So if we can abstract that complexity that is a great advantage IMO

app/tests/e2e/test_verifier.py
```diff
@@ -182,6 +185,7 @@ async def test_reject_proof_request(
         bob_and_alice_connection["alice_connection_id"],
         protocol_version=ProofRequestProtocolVersion.v10.value,
     )
+    time.sleep(3)
```
TimoGlastra, line 188: Ugh we reallyyyy need to get rid of those timeouts..... 

Webhooks FTW

TimoGlastra (approved):
LGTM! :)

app/tests/e2e/test_verifier.py
morrieinmaas: yeah will do that asap. Agreed. Slightly annoying.

morrieinmaas: Just added this because without it worked about half of the time and errored out otherwise

TimoGlastra: Yeah it's necessary at the moment 👍

PR #196, opened 2022-01-12
[WIP] Feat/155/webhook models

### LINKED ISSUE #155: [webhooks] add simple webhook models that are emitted when ACA-Py emits events
## Credentials

State should be issue cred v2 state. v1 state should be mapped to v2 state

### ACA-Py Topics

- [ ] `issue_credential` - prefix `v1-` to credential_id
- [ ] `issue_credential_v2_0` - prefix `v2-` to credential_id

```jsonc
{
   "topic": "credentials",
   "payload": {
        "credential_id": "v1-xxxx-xxxx-xxx-xxxx", // or v2-xxxx
         "state": "issue-credential-v2-state"
    }
}
```

## Proofs

State should be present proof v2 state. v1 state should be mapped to v2 state

### ACA-Py Topics

- [ ] `present_proof` - prefix `v1-` to proof_id
- [ ] `present_proof_v2_0` - prefix `v2-` to proof_id

```jsonc
{
   "topic": "proofs",
   "payload": {
        "proof_id": "v1-xxxx-xxxx-xxx-xxxx", // or v2-xxxx
         "state": "present-proof-v2-state"
    }
}
```

### ACA-Py Topics

- [ ] `connections` - prefix `v1-` to proof_id
- [ ] `present_proof_v2_0` - prefix `v2-` to proof_id

```jsonc
{
   "topic": "connections",
   "payload": {
        "connection_id": "0023-xxxx-xxxx-xxxx-xxxx",
         "state": "oob-state"
    }
}
```

### Basic Messages

```jsonc
{
   "topic": "basic-messages",
   "payload": {
         "content": "hello"
    }
}
```

morrieinmaas:
closes #155

### DIFF
--- a/webhooks/main.py
+++ b/webhooks/main.py
@@ -1,5 +1,6 @@
 from fastapi import FastAPI, Request, Depends, APIRouter
-from dependency_injector.wiring import Container, inject, Provide
+from dependency_injector.wiring import inject, Provide
+from containers import Container
 from fastapi_websocket_pubsub import PubSubEndpoint
 from starlette.websockets import WebSocket
 
--- a/webhooks/main.py
+++ b/webhooks/main.py
@@ -1,3 +1,4 @@
+from typing import List
 from fastapi import FastAPI, Request, Depends, APIRouter
 from dependency_injector.wiring import inject, Provide
 from containers import Container
@@ -5,7 +6,7 @@ from fastapi_websocket_pubsub import PubSubEndpoint
 from starlette.websockets import WebSocket
 
 from containers import Container as RedisContainer
-from services import Service
+from services import Service, TopicItem
 
 import logging
 from pprint import pformat
@@ -27,9 +28,8 @@ app.include_router(router)
 @inject
 async def index(
     topic: str, service: Service = Depends(Provide[Container.service])
-) -> dict:
-    value = await service.get_all_by_topic(f"{topic}")
-    return {f"{topic}": value}
+) -> List[TopicItem]:
+    return await service.get_all_by_topic(topic)
 
 
 @app.api_route("/{wallet_id}}")
@@ -79,6 +79,14 @@ async def websocket_rpc_endpoint(websocket: WebSocket):
 #     while True:
 #         data = websocket.receive_text()
 #         getattr(log, LOG_LEVEL)(f"\n{data}")
+
+
+@app.api_route("/")
+@inject
+async def index(service: Service = Depends(Provide[Container.service])):
+    value = await service.process()
+    return {"result": value}
+
 
 app.include_router(router)
 
--- a/webhooks/models.py
+++ b/webhooks/models.py
@@ -1,0 +1,155 @@
+from typing import Any, Optional
+
+from typing_extensions import Literal
+from pydantic import BaseModel
+
+
+class ConnectionsHook(BaseModel):
+    accept: Literal["manual", "auto"]
+    connection_id: str
+    connection_protocol: Literal["connections/1.0", "didexchange/1.0"]
+    created_at: str
+    invitation_key: Optional[str]
+    invitation_mode: Literal["once", "multi", "static"]
+    invitation_msg_id: Optional[str] = None
+    my_did: Optional[str]
+    request_id: Optional[str]
+    rfc23_state: Literal[
+        "start",
+        "invitation-received",
+        "request-sent",
+        "response-received",
+        "abandoned",
+        "completed",
+        "start",
+        "invitation-sent",
+        "request-received",
+        "response-sent",
+        "abandoned",
+        "completed",
+    ]
+    routing_state: Optional[str]
+    state: Literal[
+        "init", "invitation", "request", "response", "active", "inactive", "error"
+    ]
+    their_label: Optional[str] = None
+    their_role: Literal["invitee", "requester", "inviter", "responder"]
+    updated_at: str
+
+
+class BasicMessagesHook(BaseModel):
+    connection_id: str
+    content: str
+    message_id: str
+    sent_time: str
+    state: Literal["received"]
+
+
+class ProofsHookV1(BaseModel):
+    connection_id: Optional[str]
+    created_at: str
+    initiator: Literal["self", "external"]
+    presentation_exchange_id: str
+    presentation_request: dict
+    presentation_request_dict: dict
+    role: Literal["prover", "verifier"]
+    state: Literal[
+        "proposal_sent",
+        "proposal_received",
+        "request_sent",
+        "request_received",
+        "presentation_sent",
+        "presentation_received",
+        "verified",
+    ]
+    thread_id: str
+    trace: bool
+    updated_at: str
+
+
+class ProofsHookV2(BaseModel):
+    connection_id: Optional[str]
+    created_at: str
+    initiator: Literal["self", "external"]
+    pres_ex_id: str
+    pres_request: Optional[dict]
+    role: Literal["prover", "verifier"]
+    state: Literal[
+        "proposal-sent",
+        "proposal-received",
+        "request-sent",
+        "request-received",
+        "presentation-sent",
+        "presentation-received",
+        "verified",
+    ]
+    thread_id: str
+    trace: bool
+    updated_at: str
+
+
+class CredententialHookV2(BaseModel):
+    auto_issue: Optional[bool]
+    auto_offer: Optional[bool]
+    auto_remove: Optional[bool]
+    by_format: dict
+    connection_id: Optional[str]
+    created_at: str
+    cred_ex_id: str
+    cred_offer: Optional[dict]
+    cred_preview: Optional[dict]
+    cred_proposal: Optional[dict]
+    initiator: Literal["self", "external"]
+    role: Literal["issuer", "holder"]
+    state: Optional[
+        Literal[
+            "proposal-sent",
+            "proposal-received",
+            "offer-sent",
+            "offer-received",
+            "request-sent",
+            "request-received",
+            "credential-issued",
+            "credential-received",
+            "done",
+        ]
+    ]
+    thread_id: Optional[str]
+    trace: bool
+    updated_at: str
+
+
+class CredentialHookV1(BaseModel):
+    auto_issue: Optional[bool]
+    auto_offer: Optional[bool]
+    auto_remove: Optional[bool]
+    connection_id: Optional[str]
+    created_at: str
+    credential_definition_id: str
+    credential_exchange_id: Optional[str]
+    credential_offer: Optional[dict]
+    credential_proposal_dict: Optional[dict]
+    initiator: Literal["self", "external"]
+    role: Literal["issuer", "holder"]
+    schema_id: Optional[str]
+    state: Optional[
+        Literal[
+            "proposal_sent",
+            "proposal_received",
+            "offer_sent",
+            "offer_received",
+            "request_sent",
+            "request_received",
+            "credential_issued",
+            "credential_received",
+            "done",
+        ]
+    ]
+    thread_id: Optional[str]
+    trace: bool
+    updated_at: str
+
+
+class TopicItem(BaseModel):
+    topic: str
+    payload: Any
--- a/webhooks/services.py
+++ b/webhooks/services.py
@@ -1,14 +1,56 @@
 """Services module."""
 
+import json
+from json.decoder import JSONDecodeError
+from typing import List, Any
+
 from aioredis import Redis
+from pydantic import BaseModel
+
+
+class TopicItem(BaseModel):
+    topic: str
+    payload: Any
 
 
 class Service:
     def __init__(self, redis: Redis) -> None:
         self._redis = redis
 
-    async def add_topic_entry(self, topic: str, hook: str):
+    def __is_proof(self, topic: str) -> bool:
+        proofs_identifiers = ["present_proof", "present_proof_v2_0"]
+        return topic in proofs_identifiers
+
+    def __is_credential(self, topic: str) -> bool:
+        credential_identifiers = ["issue_credential", "issue_credential_v2_0"]
+        return topic in credential_identifiers
+
+    def __deserialise(self, data: bytes):
+        try:
+            return (
+                data.decode("utf-8")
+                .replace("\\\\", "\\")
+                .replace("\n", "")
+                .replace("'", '"')
+                .replace("False", json.dumps(False))
+                .replace("True", json.dumps(True))
+            )
+        except JSONDecodeError:
+            return data
+
+    # def _proof_hook_versioned(self, hook: bytes, topic: str) -> dict:
+    #     hook_dict = self._hook_dict(hook=hook)
+    #     hook_dict["credential_id"] = topic + hook_dict["credential_id"]
+    #     return hook_dict
+
+    #     hook_dict = json.loads(hook.decode("utf-8").replace("'", '"'))
+    #     hook_dict["proof_"]
+    #     return json.loads(hook.decode("utf-8").replace("'", '"'))
+
+    async def add_topic_entry(self, topic: str, hook: bytes):
         return await self._redis.sadd(topic, hook)
 
-    async def get_all_by_topic(self, topic: str) -> str:
-        return await self._redis.smembers(topic)
+    async def get_all_by_topic(self, topic: str) -> List[TopicItem]:
+        data = await self._redis.smembers(topic)
+        data = [self.__deserialise(val) for val in data]
+        return [TopicItem(topic=topic, payload=json.loads(payload)) for payload in data]
--- a/webhooks/services.py
+++ b/webhooks/services.py
@@ -2,30 +2,62 @@
 
 import json
 from json.decoder import JSONDecodeError
-from typing import List, Any
+from typing import List, Union
+import re
 
 from aioredis import Redis
-from pydantic import BaseModel
 
-
-class TopicItem(BaseModel):
-    topic: str
-    payload: Any
+from models import (
+    ConnectionsHook,
+    CredententialHookV2,
+    CredentialHookV1,
+    ProofsHookV1,
+    ProofsHookV2,
+    BasicMessagesHook,
+    TopicItem,
+)
 
 
 class Service:
     def __init__(self, redis: Redis) -> None:
         self._redis = redis
 
-    def __is_proof(self, topic: str) -> bool:
+    def _is_proof(self, topic: str) -> bool:
         proofs_identifiers = ["present_proof", "present_proof_v2_0"]
         return topic in proofs_identifiers
 
-    def __is_credential(self, topic: str) -> bool:
+    def _is_credential(self, topic: str) -> bool:
         credential_identifiers = ["issue_credential", "issue_credential_v2_0"]
         return topic in credential_identifiers
 
-    def __deserialise(self, data: bytes):
+    def _proof_hook_versioned(self, item: dict) -> dict:
+        if "pres_ex_id" in item:
+            item["pres_ex_id"] = "v2-" + item["pres_ex_id"]
+            item = ProofsHookV2(**item)
+        elif "presentation_exchange_id" in item:
+            item["presentation_exchange_id"] = "v1-" + item["presentation_exchange_id"]
+            item = ProofsHookV1(**item)
+        return item
+
+    def _credential_hook_versioned(self, item: dict) -> dict:
+        if "cred_ex_id" in item:
+            item["cred_ex_id"] = "v2-" + item["cred_ex_id"]
+            item = CredententialHookV2(**item)
+        elif "credential_exchange_id" in item:
+            item["credential_exchange_id"] = "v1-" + item["credential_exchange_id"]
+            item = CredentialHookV1(**item)
+        return item
+
+    def _version_connections(self, item: dict) -> dict:
+        if item["connection_protocol"] == "didexchange/1.0":
+            item["connection_id"] = "0023-" + item["connection_id"]
+        elif item["connection_protocol"] == "connections/1.0":
+            item["connection_id"] = "0016-" + item["connection_id"]
+            print(item["connection_id"])
+        item = ConnectionsHook(**item)
+        return item
+
+    def _deserialise(self, data: bytes) -> Union[str, dict]:
         try:
             return (
                 data.decode("utf-8")
@@ -36,21 +68,49 @@ class Service:
                 .replace("True", json.dumps(True))
             )
         except JSONDecodeError:
-            return data
+            return {"invalid JSON data": data}
 
-    # def _proof_hook_versioned(self, hook: bytes, topic: str) -> dict:
-    #     hook_dict = self._hook_dict(hook=hook)
-    #     hook_dict["credential_id"] = topic + hook_dict["credential_id"]
-    #     return hook_dict
+    def _to_json(self, data: List[bytes]) -> List[dict]:
+        data = [self._deserialise(val) for val in data]
+        try:
+            data = [json.loads(d) for d in data]
+        except JSONDecodeError:
+            # A multiline string has occured that has been split in
+            # an invalid way wrt JSON format
 
-    #     hook_dict = json.loads(hook.decode("utf-8").replace("'", '"'))
-    #     hook_dict["proof_"]
-    #     return json.loads(hook.decode("utf-8").replace("'", '"'))
+            # Remove multiple spaces which occur when error msg is split into multiple lines
+            # turn off black formatter here because it replaces the regex
+            # fmt: off
+            data = [re.sub('\s+',' ', d) for d in data]
+            # Remove \" \" from the multiline error message because that is invalid json
+            data = [d.replace('\" \"', ' ') for d in data]
+            # fmt: on
+            data = [json.loads(d) for d in data]
+        return data
+
+    def _to_model(self, topic: str, data: List[dict]) -> List[TopicItem]:
+        data = self._to_json(data=data)
+
+        if self._is_proof(topic):
+            data = [self._proof_hook_versioned(d) for d in data]
+        elif self._is_credential(topic):
+            data = [self._credential_hook_versioned(d) for d in data]
+        elif topic == "connections":
+            data = [self._version_connections(d) for d in data]
+        elif topic == "basicmessages":
+            data = [BasicMessagesHook(**d) for d in data]
+        return (
+            [TopicItem(topic=topic, payload=payload) for payload in data]
+            if data
+            else []
+        )
 
     async def add_topic_entry(self, topic: str, hook: bytes):
         return await self._redis.sadd(topic, hook)
 
     async def get_all_by_topic(self, topic: str) -> List[TopicItem]:
+
         data = await self._redis.smembers(topic)
-        data = [self.__deserialise(val) for val in data]
-        return [TopicItem(topic=topic, payload=json.loads(payload)) for payload in data]
+
+        items = self._to_model(topic=topic, data=data)
+        return items
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -92,33 +92,33 @@ services:
       - host
     env_file:
       - environments/yoma-multitenant/fastapi.default.env
-  # yoma-webhooks-web:
-  #   build:
-  #     context: .
-  #     dockerfile: ./dockerfiles/webhooks/Dockerfile
-  #   ports:
-  #     - "3010:3010"
-  #   volumes:
-  #     - ./webhooks:/webhooks
-  #   networks:
-  #     - yoma-multitenant
-  #     - yoma-ga
-  #     - host
-  #   env_file:
-  #     - environments/webhooks/fastapi.webhooks.env
-  #   environment:
-  #     - REDIS_URL=redis://wh-redis
-  # wh-redis:
-  #   image: redis:latest
-  #   restart: always
-  #   ports:
-  #     - "6379:6379"
-  #   volumes:
-  #     - $PWD/data:/data
-  #   networks:
-  #     - host
-  #   env_file:
-  #     - environments/webhooks/redis.env
+  yoma-webhooks-web:
+    build:
+      context: .
+      dockerfile: ./dockerfiles/webhooks/Dockerfile
+    ports:
+      - "3010:3010"
+    volumes:
+      - ./webhooks:/webhooks
+    networks:
+      - yoma-multitenant
+      - yoma-ga
+      - host
+    env_file:
+      - environments/webhooks/fastapi.webhooks.env
+    environment:
+      - REDIS_URL=redis://wh-redis
+  wh-redis:
+    image: redis:latest
+    restart: always
+    ports:
+      - "6379:6379"
+    volumes:
+      - $PWD/data:/data
+    networks:
+      - host
+    env_file:
+      - environments/webhooks/redis.env
   yoma-multitenant-wallets-db:
     image: postgres:11
     container_name: yoma-multitenant-wallets-db
@@ -227,6 +227,7 @@ volumes:
   webserver-cli:
   webserver-ledger:
   nodes-data:
+  data:
 
 networks:
   yoma-multitenant:
--- a/dockerfiles/webhooks/Dockerfile
+++ b/dockerfiles/webhooks/Dockerfile
@@ -1,4 +1,5 @@
-FROM tiangolo/uvicorn-gunicorn-fastapi:python3.7
+# Must be version 3.7 for now - bug in higher versions similar to https://github.com/tiangolo/fastapi/issues/2123
+FROM tiangolo/uvicorn-gunicorn-fastapi:python3.7 
 
 COPY webhooks /webhooks
 
--- a/environments/yoma-ga/aca-py-agent.default.env
+++ b/environments/yoma-ga/aca-py-agent.default.env
@@ -26,7 +26,7 @@ ACAPY_OUTBOUND_TRANSPORT=ws
 ACAPY_OUTBOUND_TRANSPORT=http
 ACAPY_ADMIN=[0.0.0.0,3021]
 ACAPY_ENDPOINT=http://yoma-ga-agent:3020
-# ACAPY_WEBHOOK_URL=http://yoma-webhooks-web:3010
+ACAPY_WEBHOOK_URL=http://yoma-webhooks-web:3010
 # ACAPY_WEBHOOK_URL=ws://yoma-webhooks-web:3010
 ACAPY_ADMIN_API_KEY=adminApiKey
 
@@ -50,7 +50,7 @@ ACAPY_GENESIS_URL=http://ledger-browser:8000/genesis
 # TODO: enable seed for yoma agent, the setup should be initialized
 # with yoma having an endorser did. From there we can build the rest
 # of the ecosystem
-#ACAPY_WALLET_SEED=00000000000000000000000YOMA1Any1
+# ACAPY_WALLET_SEED=00000000000000000000000YOMA1Any1
 
 ACAPY_PUBLIC_INVITES=true
 
--- a/environments/yoma-multitenant/aca-py-agent.default.env
+++ b/environments/yoma-multitenant/aca-py-agent.default.env
@@ -26,7 +26,7 @@ ACAPY_OUTBOUND_TRANSPORT=ws
 ACAPY_OUTBOUND_TRANSPORT=http
 ACAPY_ADMIN=[0.0.0.0,3021]
 ACAPY_ENDPOINT=http://yoma-multitenant-agent:3020
-# ACAPY_WEBHOOK_URL=http://yoma-webhooks-web:3010
+ACAPY_WEBHOOK_URL=http://yoma-webhooks-web:3010
 # ACAPY_WEBHOOK_URL=ws://yoma-webhooks-web:3010
 ACAPY_ADMIN_API_KEY=adminApiKey
 
--- a/manage
+++ b/manage
@@ -55,7 +55,7 @@ EOF
 # Default Settings:
 # -----------------------------------------------------------------------------------------------------------------
 DEFAULT_CONTAINERS="yoma-trust-registry yoma-multitenant-wallets-db yoma-multitenant-agent tails-server ledger-nodes ledger-browser
-                    yoma-ga-wallets-db yoma-ga-agent yoma-multitenant-web yoma-ga-web"
+                    yoma-ga-wallets-db yoma-ga-agent yoma-multitenant-web yoma-ga-web wh-redis yoma-webhooks-web"
 # -----------------------------------------------------------------------------------------------------------------
 # Functions:
 # -----------------------------------------------------------------------------------------------------------------
--- a/webhooks/requirements.txt
+++ b/webhooks/requirements.txt
@@ -2,5 +2,7 @@ aioredis~=2.0.0
 dependency_injector~=4.37.0
 fastapi==0.65.2
 fastapi_websocket_pubsub~=0.1.21
+typing_extensions
+aries-cloudcontroller~=0.4.1
 starlette
 uvicorn[standard]==0.15.0
--- a/webhooks/requirements.txt
+++ b/webhooks/requirements.txt
@@ -2,5 +2,5 @@ aioredis~=2.0.0
 dependency_injector~=4.37.0
 fastapi==0.65.2
 fastapi_websocket_pubsub~=0.1.21
-starlette==0.16.0
+starlette
 uvicorn[standard]==0.15.0


morrieinmaas:
@TimoGlastra No need for feedback just yet - still WIP.

morrieinmaas:
@TimoGlastra ready for first review. I expect there'll be a lot of remarks. Maybe some etxra info ahead:

THis now implements both http and pubsub (over websockets). The fastapi (tests) use the http handle. That is because injecting a synchronous websocket listener in the async runtime of fastapi and how things are set up was a bit of a challenge aka it got really hacky and I abandoned that road. For a client (app) however listening to pubsub/websocket will possibly or even likely be more convenient. In that case I reckon we should directly call the webhook handler (container) and not go via the cloudapi main app. Also people have a choice then between using http polling or websocket. I'm under the impression that for some folks websockets will be nicer but polling is easier to implement and what most people have done before. 

Happy to take out the pubsub mechanism altogether if you want though - for whatever reason e.g. there should only be one way of doing things.

app/generic/webhooks.py
```diff
@@ -0,0 +1,40 @@
...
+from app.generic.models import ProofsHook, ConnectionsHook, CredentialsHooks, TopicItem
+
+logger = logging.getLogger(__name__)
+
+router = APIRouter(prefix="/webhooks", tags=["webhooks"])
+
+
+@router.get("/{topic}")
+async def get_webhooks_for_wallet_by_topic(
+    topic: topics = "connections",
```
TimoGlastra, line 22: Thinking of a logical default topic, I can't think of one. Why connections over credentials, etc...? Maybe we should make the topic value required?

app/generic/webhooks.py
```diff
@@ -0,0 +1,40 @@
...
+    belongign to the wallet.
+
+    Returns:
+    ---------
+    List of webhooks belonging to the wallet
+    """
+    if "authorization" not in aries_controller.client.headers:
+        hooks = get_hooks_per_topic_admin(client=aries_controller, topic=topic)
+    else:
+        hooks = get_hooks_per_topic_per_wallet(client=aries_controller, topic=topic)
```
TimoGlastra, line 38: I've been reworking the auth in my tenant branch. It will be useful to use that here (just posting this as a reminder to myself to update)

app/generic/webhooks.py
```diff
@@ -0,0 +1,40 @@
...
+
+    This implicitly extracts the wallet ID and return only items
+    belongign to the wallet.
+
+    Returns:
+    ---------
+    List of webhooks belonging to the wallet
+    """
+    if "authorization" not in aries_controller.client.headers:
+        hooks = get_hooks_per_topic_admin(client=aries_controller, topic=topic)
```
TimoGlastra, line 36: Will this return all webhooks for all wallets, or only webhooks that are not related to a subwallet?

app/tests/e2e/test_connections.py
```diff
@@ -39,6 +43,10 @@ async def test_accept_invitation(
     )
     connection_record = accept_response.json()
 
+    assert check_webhook_state(
+        client=alice_member_client, desired_state={"state": "active"}
+    )
```
TimoGlastra, line 48: YES!! this makes the tests so much better.

Nit: this check should be placed below the next two asserts as it comes after request-sent

app/tests/e2e/test_connections.py
```diff
@@ -109,32 +108,13 @@ async def test_delete_connection(
 async def test_bob_and_alice_connect(
     bob_member_client: AsyncClient,
     alice_member_client: AsyncClient,
+    bob_and_alice_connection: BobAliceConnect,
```
TimoGlastra, line 111: I'm not sure we should use a fixture here. What are we trying to test? If we're testing the connection process we should write out the connection process I think.

app/tests/e2e/test_connections.py
```diff
@@ -109,32 +108,13 @@ async def test_delete_connection(
...
-    time.sleep(5)
-
-    bob_connection = (
-        await bob_member_client.get(f"/generic/connections/{bob_connection_id}")
-    ).json()
-    alice_connection = (
-        await alice_member_client.get(f"/generic/connections/{alice_connection_id}")
-    ).json()
+    time.sleep(2)
+    bob_connection = (await bob_member_client.get(f"/generic/connections")).json()[0]
```
TimoGlastra, line 114: Technically there could be more than one connection, so maybe we can retreive the connection by id and then do the state check?

app/tests/e2e/test_connections.py
TimoGlastra: Shouldn't we check which record id / topic we want to check the state for? Theoretically if there is a connection record with state=active this will succeed right?

app/tests/e2e/test_connections.py
```diff
@@ -109,32 +108,13 @@ async def test_delete_connection(
...
-    # FIXME: listen to webhooks once available
-    time.sleep(5)
-
-    bob_connection = (
-        await bob_member_client.get(f"/generic/connections/{bob_connection_id}")
-    ).json()
-    alice_connection = (
-        await alice_member_client.get(f"/generic/connections/{alice_connection_id}")
-    ).json()
+    time.sleep(2)
```
TimoGlastra, line 113: If we write out the test here and use the `assert_webhook_state` method we can wait for the expected webhook event and ditch the sleep (we should avoid those evil sleeps 👿  )

app/tests/e2e/test_fixtures.py
```diff
@@ -85,7 +86,12 @@ async def credential_exchange_id(
     credential_exchange_id = credential_exchange["credential_id"]
     assert credential_exchange["protocol_version"] == "v1"
 
-    time.sleep(5)
+    assert check_webhook_state(
```
TimoGlastra, line 89: Ah I see you added an `topic` option. For completeness record_id would also be good!

app/tests/e2e/test_issuer.py
```diff
@@ -91,7 +100,11 @@ async def test_get_records(alice_member_client: AsyncClient):
 async def test_send_credential_request(
     bob_member_client: AsyncClient, credential_exchange_id: str
 ):
-    time.sleep(10)
+    assert check_webhook_state(
```
TimoGlastra, line 103: I think we can fix these tests now right? With auto-accept disabled and by listening to webhook events?

app/tests/e2e/test_issuer.py
```diff
@@ -111,7 +124,11 @@ async def test_store_credential(
 ):
     # TODO check for the correct response when state is credential_received
     # We can't complete this with auto accept enabled
-    time.sleep(5)
```
TimoGlastra, line 114: Same with this tests. It doesn't make a lot of sense atm. bob sends an offer, so it makes sense it can't store the credential.

app/tests/e2e/test_verifier.py
```diff
@@ -121,12 +122,18 @@ async def test_accept_proof_request(
...
     proof_request_v1.connection_id = bob_and_alice_connection["alice_connection_id"]
     proof_dict["connection_id"] = bob_and_alice_connection["alice_connection_id"]
-    time.sleep(3)
+
     proof_req_res = await alice_member_client.post(
         BASE_PATH + "/send-request",
         json=proof_request_v1.dict(),
     )
 
+    assert check_webhook_state(
```
TimoGlastra, line 131: Same here

app/facades/webhooks.py
```diff
@@ -0,0 +1,71 @@
...
+    "issuer_cred_rev",
+    "issue_credential_v2_0",
+    "issue_credential_v2_0_indy",
+    "issue_credential_v2_0_dif",
+    "present_proof",
+    "revocation_registry",
+]
+
+
+def get_wallet_id_from_client(client: Union[AcaPyClient, AsyncClient]) -> str:
```
TimoGlastra, line 29: A problem with this approach is that we can't verify if the JWT is valid. For all other requests we're calling ACA-Py which will verify whether the JWT is valid. 

I can just create a JWT with a wallet id and you can retrieve all webhooks for a specific wallet: https://jwt.io/#debugger-io?token=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ3YWxsZXRfaWQiOiJteV93YWxsZXRfaWQifQ.9ALYlsIGqjfzP1s2EzUa0tCIjWWPElAyKmlsrPIzeOw

You still need the base adminApiKey, so not that much of a problem in this case, but not sure if we're creating a change for security issues here

app/tests/util/webhooks.py
```diff
@@ -0,0 +1,55 @@
+import time
+import os
+from typing import List, Literal
+
+from httpx import AsyncClient
+import httpx
+
+from app.tests.util.string import get_wallet_id_from_JWT
+
+WEBHOOKS_URL = os.getenv("WEBHOOKS_URL", "http://localhost:3010")
```
TimoGlastra, line 10: We have a constants file for tests that contain env variables for tests: https://github.com/didx-xyz/aries-cloudapi-python/blob/development/app/tests/util/constants.py

app/tests/util/webhooks.py
```diff
@@ -0,0 +1,55 @@
...
+
+topics = Literal[
```
TimoGlastra, line 12: This is redeclaring the topic from the webhook facade, maybe we can reuse that list to avoid inconsistencies in the future?

app/tests/util/webhooks.py
```diff
@@ -0,0 +1,55 @@
...
+    "issue_credential_v2_0_dif",
+    "present_proof",
+    "revocation_registry",
+]
+
+
+def check_webhook_state(
+    client: AsyncClient,
+    desired_state: dict = None,
+    topic: topics = "connections",
```
TimoGlastra, line 30: Ah I see you also used the `connections` topic as default here. I think if we make the property required the api of the method would be more clear (I don't have to look in the implementation what the topic is if I don't provide one)

webhooks/clients.py
```diff
@@ -28,13 +31,16 @@ async def main():
...
         "revocation_registry",
     ]
-    client = PubSubClient([], callback=on_events)
+    client = PubSubClient([*topics], callback=on_events)
 
-    async def on_data(data, topic):
-        print(f"{topic}:\n", data)
+    # You can also register it using the commented code below
+    # async def on_data(data, topic):
+    #     print(f"{topic}:\n", data)
```
TimoGlastra, line 38:
```suggestion
```

webhooks/clients.py
```diff
@@ -28,13 +31,16 @@ async def main():
...
 
-    [client.subscribe(topic, on_data) for topic in topics]
-    client.start_client(f"ws://0.0.0.0:{PORT}/pubsub")
+    # [client.subscribe(topic, on_data) for topic in topics]
+    # client.subscribe("connections", on_data)
```
TimoGlastra, line 41:
```suggestion
```

webhooks/clients.py
```diff
@@ -28,13 +31,16 @@ async def main():
         "present_proof",
         "revocation_registry",
```
TimoGlastra, line 32: What if events are emitted for other than these topics? I think we're missing a few related to present proof v2 (and maybe others)

webhooks/clients.py
```diff
@@ -4,13 +4,16 @@
...
 PORT = os.getenv("PORT", "3010")
+URL = os.getenv("WEBHOOKS_URL", "127.0.0.1")
 
 
 async def on_events(data, topic):
-    print(f"{topic}: {data}")
+    print(f"{topic}:\n{data}")
 
 
 async def main():
```
TimoGlastra, line 19: What does this method do? Does it subscribe to ACA-Py webhooks? Or is it for others to subscribe to events from the CloudAPI?

webhooks/main.py
```diff
@@ -64,20 +85,11 @@ async def topic_wallet(
...
-# async def websocket_topics(websocket: WebSocket):
-#     await websocket.accept()
-#     while True:
-#         data = websocket.receive_text()
-#         getattr(log, LOG_LEVEL)(f"\n{data}")
+# # Example for broadcasting from eg Redis
+# @router.websocket("/pubsub")
+# async def websocket_rpc_endpoint(websocket: WebSocket):
+#     async with endpoint.broadcaster:
+#         await endpoint.main_loop(websocket)
```
TimoGlastra, line 92: Is the pubsub functionality removed?

webhooks/main.py
```diff
@@ -45,9 +58,17 @@ async def wallet_root(
 async def topic_root(
     topic, request: Request, service: Service = Depends(Provide[Container.service])
 ):
-    payload = pformat(await request.json())
-    await service.add_topic_entry(str(topic), str(payload))
+    payload = await request.json()
+    wallet_id = {"wallet_id": (request.headers)["x-wallet-id"]}
```
TimoGlastra, line 62:
```suggestion
    wallet_id = {"wallet_id": request.headers["x-wallet-id"]}
```

TimoGlastra: Will this work if wallet_id is not defined?

webhooks/main.py
```diff
@@ -45,9 +58,17 @@ async def wallet_root(
...
+    payload.update(wallet_id)
+    payload = pformat(payload)
+    # redistribute by topic
     await endpoint.publish(topics=[topic], data=payload)
+    # Add data to redis
+    await service.add_topic_entry(str(topic), str(payload))
+    # redistribute per wallet
+    wallet_data = service.get_all_for_topic_by_wallet_id
```
TimoGlastra, line 70: Shouldn't this method be called?

webhooks/models.py
```diff
@@ -0,0 +1,170 @@
+from typing import Any, Optional
+from typing_extensions import Literal
+from pydantic import BaseModel, ValidationError
+
+
+class HookBase(BaseModel):
+    wallet_id: Optional[str]
+
+
+class ConnectionsHook(HookBase):
```
TimoGlastra, line 10: This doesn't follow the examples from the webhooks issue: https://github.com/didx-xyz/aries-cloudapi-python/issues/155

What is the reason you went with the full payloads? 

The webhook models are currently not matching with the models returned from the CloudAPI. I think it would be good to keep those as closely aligned as possible as now the user needs to handle two structures for the same data payload. 

Maybe we can keep the raw webhooks in the webhooks app so we can consume the full webhook in the CloudApi, but we transform it to the CloudApi models before sending it?

webhooks/models.py
```diff
@@ -0,0 +1,170 @@
...
+    thread_id: Optional[str]
+    trace: bool
+    updated_at: str
+    credential_exchange_id: Optional[str]
+    credential_definition_id: Optional[str]
+    credential_proposal: Optional[dict]
+    credential_offer: Optional[dict]
+
+
+def to_credentential_hook_model(item: dict) -> CredentialsHooks:
```
TimoGlastra, line 135: Would be good to reuse the app logic here. We already have methods that transform v1 and v2 ACA-Py credential / proof records to the CloudAPI model.

webhooks/services.py
```diff
@@ -1,14 +1,113 @@
...
+
+    def _deserialise(self, data: bytes) -> Union[str, dict]:
+        # Turn bytes from redis into str
+        try:
+            return (
+                data.decode("utf-8")
+                .replace("'", '"')
+                .replace("False", json.dumps(False))
+                .replace("True", json.dumps(True))
+            )
```
TimoGlastra, line 48: Why is this needed? We do a json.dumps before sending the data to redis right? Couldn't we do a json.loads here? This feels quite hacky

webhooks/services.py
```diff
@@ -1,14 +1,113 @@
...
         return await self._redis.sadd(topic, hook)
 
-    async def get_all_by_topic(self, topic: str) -> str:
-        return await self._redis.smembers(topic)
+    async def get_all_by_topic(self, topic: str) -> List[TopicItem]:
+        # Get the data from redis queue
+        data = await self._redis.smembers(topic)
+        # Transform it to TopicItems
+        items = self._to_topic_item(topic=topic, data=data)
+        return items
```
TimoGlastra, line 101: Shouldn't we include paging here? What if we have a loooot of webhooks events?

webhooks/services.py
```diff
@@ -1,14 +1,113 @@
...
+
+    async def get_all_for_topic_by_wallet_id(
```
TimoGlastra, line 103: Same here

app/facades/webhooks.py
```diff
@@ -0,0 +1,71 @@
...
+from typing import Union
+import os
+from typing import List, Literal
+from aries_cloudcontroller import AcaPyClient
+
+from httpx import AsyncClient, get, HTTPError
+
+
+WEBHOOKS_URL = os.getenv("WEBHOOKS_URL", "http://yoma-webhooks-web:3010")
+ADMIN_API_KEY = os.getenv("ACAPY_ADMIN_API_KEY", "adminApiKey")
```
TimoGlastra, line 12: We have a constant file to keep values

app/facades/webhooks.py
```diff
@@ -0,0 +1,71 @@
...
+        raise e from e
+
+
+def get_hooks_per_topic_admin(client: AcaPyClient, topic: topics) -> List:
+    """
+    Gets all webhooks for all wallets by topic (default="connections")
+    """
+    try:
+        # Ensure admin key is present
+        assert client.client.headers["x-api-key"] == ADMIN_API_KEY
```
TimoGlastra, line 64: With mulitple ACA-Py instances there are multiple admin api keys right?

app/facades/webhooks.py
```diff
@@ -0,0 +1,71 @@
...
+        # Ensure it's not a wallet/tenant
+        assert "authorization" not in client.client.headers
+        hooks = (get(f"{WEBHOOKS_URL}/{topic}")).json()
+        # Only return the first 100 hooks to prevent OpenAPI interface from crashing
+        return hooks[:100] if hooks else []
```
TimoGlastra, line 69: What if I want to retrieve webhook 101?

TimoGlastra (commented):
Overall looks to me @morrieinmaas. 

One thing I'm asking myself is whether we're not overcomplicating the webhooks.  Thinking of it webhooks should provide a few things:
1. Services that need updates from the CloudAPI can subscribe to webhook events. They'll most likely process webhooks in realtime and not care about them after they fired. Actions are immediate.
2. The CloudAPI needs updates from ACA-Py so it can automate certain processes. It will listen for webhook events, act directly and forget about the after it. 

For both cases the webhook content can easily be retrieved by fetching the latest record state.  Looking at other SSI implementations (Trinsic, ACA-Py, Verity) AFAIK all just emit webhook events and forget about them afterwards.

What will be the use case for someone retrieving the webhooks using the API? How does it know which webhooks it has already processed? I think it may make more sense to send webhook events directly to the user of the CloudAPI and we forget about them afterwards. Will remove a whole lot of complexity. But maybe I'm missing something here?

app/generic/webhooks.py
morrieinmaas: TBH I agree with you here and am also in favour of making topic required.

I did connections because it's sort of the first thing one establishes in any kind of flow but really not much more reason to this than that. so I'll remove the default

app/generic/webhooks.py
morrieinmaas: all webhooks stored for a topic

app/tests/e2e/test_connections.py
morrieinmaas: well spotted and good question. It is taken care of though (implicitly). the method extracts the wallet-id and gets the connection records only for that wallet. the wallet is crreated (and destroyed) per test run and stems from the fixture so there is only that connections between alice and bob. Could add checking against the connection ID though to be more explicit, maybe that's better/clearer

app/tests/e2e/test_connections.py
morrieinmaas: fair point

app/tests/e2e/test_connections.py
morrieinmaas: agreed

app/tests/e2e/test_connections.py
morrieinmaas: agreed

app/tests/e2e/test_issuer.py
morrieinmaas: yes we can - in fact I left some out here because I wanted feedback first. Wanted to avoid rewriting it all to potentially re-write it again then in case feedback was more substantial

app/facades/webhooks.py
morrieinmaas: I see what you mean, but there is two reasons why I thought that not a problem:

1. Thsi method is not exposed to the user directly and relies on extracting the token from controller/client which gets instantiated via fastapi mechanism
2. The fastapi can remain the gatekeeper (as we do know already).

If wanted I can also add a verification of the jwt?!

app/tests/util/webhooks.py
morrieinmaas: Ah yeah true. Sorry I went the old way of doing that. You;r absolutely right

webhooks/clients.py
morrieinmaas: will add

webhooks/clients.py
morrieinmaas: This clients is only an example of how to roll your own client so people can see how to use it

webhooks/main.py
morrieinmaas: no one can still use that is desired

webhooks/main.py
morrieinmaas: probably yes. let me have a look - intrigued I accidentally deleted that with undo or sth

webhooks/models.py
morrieinmaas: Yes, I took the liberty to go for the full ones because I thought this makes them more versatile and people can use them for other cases/don't have to extend them. I'm happy to make them as per issue though if you think that's and overkill. 

I'm all in favour of making the models as then ones returned from cloudapi though (as a personal preference of what I interpret as 3 options to go from this comment)

webhooks/models.py
morrieinmaas: yes. However, as far as I know I'd still have to write that code out here because sharing between the two apps/containers isn't really doable nicely - but let me know if you have a good idea though

webhooks/services.py
morrieinmaas: json.loads unfortunately doesn't cut the mustard here. That's what I tried first to no avail.

webhooks/services.py
morrieinmaas: could add paging indeed. Need to have a look how to do that

app/facades/webhooks.py
morrieinmaas: yeah i guess paging as you mentioned up is the way to go. I only added this because the OpenAPi UI crashes if if there's a lot of webhooks

morrieinmaas:
> Overall looks to me @morrieinmaas.
> 
> One thing I'm asking myself is whether we're not overcomplicating the webhooks. Thinking of it webhooks should provide a few things:
> 

I've been wondering this as well. On the other side though we can now handle the webhooks anyway we want and implement more functionality (using webhooks) that currently isn't covered without having to extend the webhooks handler every time. 


>     1. Services that need updates from the CloudAPI can subscribe to webhook events. They'll most likely process webhooks in realtime and not care about them after they fired. Actions are immediate.
> 
>     2. The CloudAPI needs updates from ACA-Py so it can automate certain processes. It will listen for webhook events, act directly and forget about the after it.
> 

IMHO there is also a third case which is the client apps (YOMA is writing in F# or sth). I was imagining eg a mobile app or or anything similar that can have a lot of connections and proofs and credentials and goes offline for a while. By exposing the webhooks this way with "memory" the app can return online and get a bunch of events they "missed" and act upon that as opposed to explicitely having to ask has a) b) or c) happened which is fetching record state for example 

> 
> For both cases the webhook content can easily be retrieved by fetching the latest record state. Looking at other SSI implementations (Trinsic, ACA-Py, Verity) AFAIK all just emit webhook events and forget about them afterwards.
> 
> What will be the use case for someone retrieving the webhooks using the API? How does it know which webhooks it has already processed? I think it may make more sense to send webhook events directly to the user of the CloudAPI and we forget about them afterwards. Will remove a whole lot of complexity. But maybe I'm missing something here?

You do have a point here and I neither agree nor disagree really. It's more a question of what it's meant to handle. I was under the impression that it's partly about not forgetting the hooks directly. Otherwise we could just decided to not use redis and redistribute the hooks straight away (which is also fine - just a matter of choice imho).

Let me know your thoughts  and thanks for all the feedback already

webhooks/clients.py
morrieinmaas: There's been (to my bewilderement) a difference in topics available and documented ([here](https://github.com/hyperledger/aries-cloudagent-python/blob/main/AdminAPI.md#administration-api-webhooks)) as well as a difference in structure between webhook received and webhook expected as per [documentation](https://github.com/hyperledger/aries-cloudagent-python/blob/main/AdminAPI.md#administration-api-webhooks)

morrieinmaas: Stupid question but where is an exhaustive list of all topics documented?

webhooks/services.py
morrieinmaas: You receive a byte stream from redis that you need to convert and the webhook itself is not always json conform so if we don't do this we get JSONDecode errors.

webhooks/models.py
morrieinmaas: I'll have a go and transform the models to be like what we have in the API already where applicable.

webhooks/services.py
morrieinmaas: I've had a look and that is not a standard feature in fastapi so I have to use some plugin or roll my own

app/facades/webhooks.py
morrieinmaas: Hm - that is a good question. I didn't think of that. Is that a realistic scenario? and how would the client handle that (when targeting an aca-py instance)? Wouldn't it have to knwo which one to use then somehow (the client I mean)? If this is a problem here, isn't that also a problem in general? Maybe we can have a quick discussion about this. I'm not quite sure

webhooks/services.py
morrieinmaas: since this will be accessed by code I think paging for human-readable friendliness is not required

app/facades/webhooks.py
morrieinmaas: This is still outstanding as I'm nto yet sure how to handle this best, but working on it.

webhooks/models.py
morrieinmaas: I've made the models shared between the two apps now.

For the validation logic, as much as I agree with this being better shared also, I am in favour of doing this as an improvement in a separate PR. That is for the cloudapi app has this logic partially as class (private) methods and further validates this against aries-cloudcontroller types that are being returned. The webhooks do this against a bytestream from Redis instead. Having had a look at it, I think this a slightly delicate matter that requires some effort that lies beyond the scope of this PR imho. Hence, I'd do it in a different PR.

morrieinmaas:
@TimoGlastra Do you have any more feedback here?

shared_models/shared_models.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+class ProofRequestProtocolVersion(Enum):
+    v10 = "v1"
+    v20 = "v2"
+
+
+class IssueCredentialProtocolVersion(Enum):
+    v1 = "v1"
+    v2 = "v2"
```
TimoGlastra, line 18: Inconsistent keys

shared_models/shared_models.py
```diff
@@ -0,0 +1,97 @@
...
+    my_did: Optional[str]
+    alias: Optional[str] = None
+    their_did: Optional[str] = None
+    their_label: Optional[str] = None
+    their_public_did: Optional[str] = None
+    updated_at: Optional[str] = None
+    error_msg: Optional[str] = None
+    invitation_key: Optional[str] = None
+    invitation_msg_id: Optional[str] = None
+    invitation_msg_id: Optional[str] = None
```
TimoGlastra, line 37: duplicate

webhooks/models.py
```diff
@@ -0,0 +1,98 @@
...
+
+class ConnectionsHook(HookBase, Connection):
+    pass
+
+
+def to_connections_model(item: dict) -> ConnectionsHook:
+    if item["connection_protocol"] == "didexchange/1.0":
+        item["connection_id"] = "0023-" + item["connection_id"]
+    elif item["connection_protocol"] == "connections/1.0":
+        item["connection_id"] = "0016-" + item["connection_id"]
```
TimoGlastra, line 20: We don't prefix the connection id in the cloud api, so to stay consistent I think we also shouldn't do it for the webhooks (they should match 1:1)

The reason we did this for credentials / proofs is that they have separate records under the hood for v1 and v2. But with connections it is the same record for didexchange and connections

webhooks/models.py
```diff
@@ -0,0 +1,98 @@
...
+    message_id: str
+    sent_time: str
+    state: Literal["received"]
+
+
+class ProofsHook(HookBase, PresentationExchange):
+    pass
+
+
+def to_proof_hook_model(item: dict) -> ProofsHook:
```
TimoGlastra, line 41: Does this match with the cloud api webhook models. We should be 1:1 consistent here, even if ACA-Py isn't

webhooks/services.py
```diff
@@ -1,14 +1,137 @@
...
+    def _proof_hook_versioned(self, item: dict) -> dict:
+        return to_proof_hook_model(item=item)
+
+    def _credential_hook_versioned(self, item: dict) -> dict:
+        return to_credentential_hook_model(item=item)
+
+    def _version_connections(self, item: dict) -> dict:
+        return to_connections_model(item=item)
+
+    def _deserialise(self, data: bytes) -> Union[str, dict]:
```
TimoGlastra, line 40: Why is this needed? Can't we store stringified json in redis so we don't have to transform python dict values to json values?

webhooks/services.py
```diff
@@ -1,14 +1,137 @@
...
+                    payload=payload.dict(),
+                )
+                for payload in data
+            ]
+            if data
+            else []
+        )
+
+    async def add_topic_entry(self, topic: str, hook: bytes) -> List[TopicItem]:
         return await self._redis.sadd(topic, hook)
```
TimoGlastra, line 103: Couldn't we transform the hook to json here?

webhooks/main.py
```diff
@@ -40,44 +54,58 @@ async def wallet_root(
...
+    service: Service = Depends(Provide[Container.service]),
 ):
-    payload = pformat(await request.json())
-    await service.add_topic_entry(str(topic), str(payload))
+    payload = await request.json()
+    wallet_id = {"wallet_id": request.headers["x-wallet-id"]}
+    payload.update(wallet_id)
+    origin_id = {"origin": origin}
+    payload.update(origin_id)
+    payload = pformat(payload)
```
TimoGlastra, line 72: Instead of doing `str(pformat(payload))` can't we do `json.dumps(payload)` and encode it to bytes? Seems less hacky to me

webhooks/services.py
```diff
@@ -1,14 +1,137 @@
...
-    async def get_all_by_topic(self, topic: str) -> str:
-        return await self._redis.smembers(topic)
+    async def get_all_by_topic(self, topic: str) -> List[TopicItem]:
+        # Get the data from redis queue
+        data = await self._redis.smembers(topic)
+        # Transform it to TopicItems
+        items = self._to_topic_item(topic=topic, data=data)
+        for d in items:
+            del d.payload["wallet_id"]
+            del d.payload["origin"]
```
TimoGlastra, line 112: If we don't add these keys to the payload itself, but rather create an object like this:

```
{
   "wallet_id": "xx",
   "origin": "xxx",
   "data": {}
}
```

Then we don't have to do the deleting of the properties before returing (can just return data) and we also avoid the possiblity of the attributes magically clashing. (the wallet record has a property named `wallet_id`)

webhooks/services.py
```diff
@@ -1,14 +1,137 @@
...
+        return items
+
+    async def get_all_for_topic_by_wallet_id(
+        self, topic: str, wallet_id: str
+    ) -> List[TopicItem]:
+        data = await self._redis.smembers(topic)
```
TimoGlastra, line 118: Don't we store the items by both wallet id and topic? How is this going to scale? If not possible to index by both, it may be better to index by wallet_id as I suspect all by topic and the filtering for wallet id will be much more that filtering all by wallet id and then filtering for topic (I suspect there will be a looooooooot of wallets, that each are not that big)

app/tests/e2e/test_connections.py
```diff
@@ -39,6 +42,16 @@ async def test_accept_invitation(
...
     connection_record = accept_response.json()
 
+    assert check_webhook_state(
+        client=alice_member_client,
+        desired_state={"state": "completed"},
+        topic="connections",
+        filter_map=FilterMap(
+            filter_key="connection_id",
+            filter_value=connection_record["connection_id"],
+        ),
```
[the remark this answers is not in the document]
morrieinmaas: sure - good idea!

webhooks/services.py
morrieinmaas: In theory we can. And I can have a go at it. However storing JSON in redis works a bit different and we'll then have to always provide a unique non-existing key to the hook that is in json. This way I could just dump it all. Happy to change it/give it a try though

TimoGlastra: Can redis differentiate between a json object encoded as string encoded as bytes and a dict encoded as string? I'm not talking about JSON directly, but rather a JSON string

webhooks/services.py
morrieinmaas: good shout

webhooks/services.py
morrieinmaas: not sure I understand your question correctly, but have a look at [this](https://developer.redis.com/howtos/redisjson/using-python/)

webhooks/services.py
morrieinmaas: As below. Yes, possible in principle

app/tests/e2e/test_connections.py
```diff
@@ -110,34 +114,41 @@ async def test_bob_and_alice_connect(
...
+    bob_connections = (await bob_member_client.get(f"/generic/connections")).json()
+    alice_connections = (await alice_member_client.get(f"/generic/connections")).json()
+    bob_connection = [
+        b for b in bob_connections if b["connection_id"] == invitation["connection_id"]
+    ]
+    alice_connection = [
+        a
+        for a in alice_connections
+        if a["connection_id"] == connection_record["connection_id"]
+    ]
```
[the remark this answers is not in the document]
morrieinmaas: my bad - will do by id

webhooks/services.py
morrieinmaas: Just a quick thought you've probably suggested the better way of doing this below either way. If we store by wallet_id this gets much simpler. Only thing then is what if we dont have a wallet id (or do we always have a wallet_id)?

webhooks/services.py
morrieinmaas: ah i remember now why I did that. Because the dtaa living in redis only has a single key and the wallet id and origin attrs get added when the webhook is received and then stored to redis under a single key. So when retrieving the data from redis we wnt that piece of informtion to be there but not in the "payload" that represents some model from aca-py or cloudapi. Hence, it's being deleted. Not sure how to get around this otherwise.

morrieinmaas: found a way nontheless i think. Will push in a little

webhooks/services.py
morrieinmaas: Just to clarify - it was only stored by topic and wallet id was added to the payload. I was using redis sets so for a key such as connections I stored (added/appended all the hooks and added "wallet_id": "sthsth" to the actual hook. then I retrieved the hooks for the topic and filtered on the data. 

Either way it's a bit of a trade-off. If we really wanna scrimp on data in favour of speed/scalability here I reckon removing storing by topic and only storing by wallet_id is an option as by topic retrieval is really just an admin/debug kind of thing

morrieinmaas: This is a bit of a tricky one. We either have to store the same data for different keys (same data for different keys is a [requested feature in redis](https://github.com/redis/redis/issues/2668)) or come up with some sort of trick.

webhooks/services.py
morrieinmaas: They can just get ditched when transforming into models

app/tests/e2e/test_connections.py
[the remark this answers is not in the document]
morrieinmaas: Also remember now whilst changing this why i did this. It's because when I use this to filter I can use filter_map.filter_key or filter_map.filter_value and be sure it works. WHere as if I have a dict and you can't be sure because you there can be multiple keys/ key value pairs etc

webhooks/services.py
morrieinmaas: only using wallet id for now then

app/tests/e2e/test_connections.py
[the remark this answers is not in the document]
morrieinmaas: another one: I kept desired_state separate because then the check can also be "sth" (key) instead of "state" key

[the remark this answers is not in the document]
morrieinmaas: I meant that if we dont want to check for "state": "completed" but "some_other_key": "some_other_value" that was imho more obvious/readable. Changed it to your proposal now though.

app/tests/e2e/test_issuer.py
```diff
@@ -89,35 +97,117 @@ async def test_get_records(alice_member_client: AsyncClient):
...
+        desired_state={"state": "offer-received"},
+        topic="issue_credential",
+    )
+
+    cred_hooks = get_hooks_per_topic_per_wallet(
+        client=alice_member_client, topic="issue_credential"
+    )
+
+    cred_hook = [h for h in cred_hooks if h["payload"]["state"] == "offer-received"][0]
+    credential_id = cred_hook["payload"]["credential_id"]
```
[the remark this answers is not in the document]
morrieinmaas: We're not storing thread_id (in the custom models)

app/tests/e2e/test_issuer.py
```diff
@@ -89,35 +97,117 @@ async def test_get_records(alice_member_client: AsyncClient):
...
+        topic="issue_credential",
+    )
+
+    response = await alice_member_client.post(f"{BASE_PATH}/{credential_id}/store")
+
+    assert check_webhook_state(
+        client=alice_member_client,
+        desired_state={"state": "credential-received"},
+        topic="issue_credential",
     )
```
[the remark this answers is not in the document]
morrieinmaas: Agreed on the latter bit. Changed that now. 

For the first paragraph:

* Bob sends a credential (offer) (automating the entire flow)
* Alice receives the offer
* as per [diagram](https://github.com/hyperledger/aries-rfcs/blob/main/features/0036-issue-credential/credential-issuance.png) alice then sends a credential request (data is correct)
* Bob replies with the credential (because we have used 'automating the entire flow - `app/generic/issuer/issuer.py::send_credential`)

=> that's how I understood this to work. Maybe I'm making a mistake here?

added a check for `credential-acked` to the test and moved the `credential-received` check in front of storing it.

app/tests/e2e/test_verifier.py
```diff
@@ -121,12 +121,18 @@ async def test_accept_proof_request(
     )
```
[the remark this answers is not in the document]
morrieinmaas: would've left that to a different PR but will have a go at it now

app/tests/e2e/test_verifier.py
morrieinmaas: Been trying to fix accept proof request for quite a while last night. Suspect there is a bug in there. Will open another issue for this.

app/generic/webhooks.py
```diff
@@ -17,6 +17,23 @@
 router = APIRouter(prefix="/webhooks", tags=["webhooks"])
 
 
+@router.get("/")
+async def get_webhooks_for_wallet(
```
TimoGlastra, line 21: This is great!

TimoGlastra (approved):
LGTM @morrieinmaas! 

Only thing that I think needs addressing is the structure of the webhooks to match the cloud api. But as discussed that can be done in a separate PR. Can you open a ticket for this?

app/tests/e2e/test_issuer.py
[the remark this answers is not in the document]
TimoGlastra: Maybe we should :-)

morrieinmaas:
> LGTM @morrieinmaas!
> 
> Only thing that I think needs addressing is the structure of the webhooks to match the cloud api. But as discussed that can be done in a separate PR. Can you open a ticket for this?

Yes can do!

PR #202, opened 2022-02-22
Make record to model logic shared

### LINKED ISSUE #200: 100% matching models for webhoosk and cloudapi
Ensure that the models for cloudapi and webhooks match as much as possible

morrieinmaas:
closes #200 

Presentation Record logic shared fully


morrieinmaas:
@TimoGlastra Any feedback welcome!

Maybe you can spot some more things that can be shared?

Think this should largely cover the shared logic and conversion to models as discussed.

webhooks/models.py
```diff
@@ -1,27 +1,30 @@
...
-    item["state"] = item.pop("rfc23_state")
-    try:
-        item = ConnectionsHook(**item)
-    except ValidationError:
-        pass
-    return item
+    conn_record = ConnRecord(**item)
+    conn_record = conn_record_to_connection(connection_record=conn_record)
+
+    merged_dicts = {**item, **conn_record.dict()}
```
TimoGlastra, line 26: Why are both the plain dict and the conn record merged here?

TimoGlastra (commented):
This looks great @morrieinmaas. On question with regards to the merged dicts.

Does this also mean the webhooks and cloudapi now 100% match?

webhooks/models.py
morrieinmaas: because the item has eg the wallet_id and oriigin fields which the conn record doesnt have.

The way this works is that the left-most gets used first (including values we dont want and have overwritten (eg moved '_' to '-')BUT then the conn record/item on the right gets used to update and override existing fields (with the values we have chanhed to the model. In essence the all the fields of the model we use get overridden the way we want and the ones we dont have in the model but specific from the webhook event come from item

And yes, they now match 100%

morrieinmaas: example:

```python
a = {'a': 1, 'b': 2, 'c':3}
b = {'b': 3, 'c':5}
c = {**a, **b}
print(c)
# gives
{'a': 1, 'b': 3, 'c':5}
```

Issue #141, opened 2021-10-25
Integrate schema endpoint of cloudapi with trustregistry


morrieinmaas:
Creating and writing a schema via the cloudapi onto the ledger should update the trustregistry as well.

- [ ] Identify all points in cloudapi that use schemas in some form or another
- [ ] Integrate the trustregistry into cloudapi such that the trustregistry is up to date e.g. deleting a schema should delete a schema from the trustregistry (as it is not possible to delete it from the ledger).
- [ ] Remove schemas endpoint as in example of previous point
- [ ] Replace get schemas list to get schemas from trustregistry instead of trying to get that from the ledger

PR #206, opened 2022-03-01
feat: integrate schema endpoint with trust registry

TimoGlastra:
Integrates the schema endpoint with the trust registry. When a schema is created, it is automatically added to the trust registry.

In addition this endpoint also removes the /list/ and update schema endpoint. I don't think they're explicitly needed. Especially the update schema adds complexity that is better handled outside of the cloud api (currently).

Fixes https://github.com/didx-xyz/aries-cloudapi-python/issues/141


morrieinmaas (approved):
LGTM

PR #208, opened 2022-03-10
feat: integrate verifier with trust registry

### LINKED ISSUE #189: integrate verifier with the trust registry
We should enforce trust registry rules in the verifier api

morrieinmaas:
closes #189


morrieinmaas:
@TimoGlastra ready for review

app/generic/verifier/verifier_utils.py
```diff
@@ -0,0 +1,171 @@
...
+        pub_did = ed25519_verkey_to_did_key(key=invitation_key)
+
+    # Try get actor from TR
+    actor = await get_actor(did=pub_did)
+    # 2. Check actor has role verifier
+    if not is_verifier(actor=actor):
+        raise CloudApiException("Could not verify prover against trust registry", 401)
+
+    # Get schema ids
+    schema_ids = await get_schema_ids(aries_controller=aries_controller, prover=prover)
```
TimoGlastra, line 84: This checks if the credentials the user has, have schema ids that  are valid in the trust request. However this doesn't actually test anything for the proof request.

The user could have a credential in its wallet with a different schema id. The thing is the verifier should only be allowed to ask for a request for one of the schemas registered in the trust registry. This can be quite hard as proof request can be defined quite dynamically.  I think we're left with two options:

1. require all restrictions on requested_attributes and request_predicates to require the schema_id to be present
2. fetch all credentials that could fulfil the request (instead of fetching _all_ credentials )and filter out credentials that don't have a correct schema id OR throw an error if any credential is in the list with a schema id not in the trust registry.

I think 1. is easiest.

app/tests/e2e/test_verifier.py
```diff
@@ -31,13 +38,18 @@ def create_send_request(connection_id: str, protocol_version: str) -> SendProofR
 async def test_send_proof_request(
     bob_and_alice_connection: BobAliceConnect,
     alice_member_client: AsyncClient,
+    schema_definition: SchemaSendResult,
+    bob_member_client: AsyncClient,
 ):
     # V1
     proof_request_v1 = create_send_request(
-        connection_id=bob_and_alice_connection["alice_connection_id"],
+        connection_id=bob_and_alice_connection["bob_connection_id"],
```
TimoGlastra, line 46: It would be good here to add a test with an actor with the role of verifier that has no public did. Or is that already added? I couldn't find any test that tests with the verifier not having a public did.

app/tests/util/member_personas.py
```diff
@@ -123,3 +138,199 @@ async def bob_and_alice_connection(
...
+    # accept invitation on alice side
+    invitation_response = (
+        await alice_member_client.post(
+            "/generic/connections/accept-invitation",
+            json={"invitation": invitation},
+        )
+    ).json()
+
+    print(f"\n\n\n\n Invitation: {invitation}\n\n\n")
+    time.sleep(10)
```
TimoGlastra, line 305: We have nice event listeners now. Sleeps make the tests unnecesarily slow

shared_models/shared_models.py
```diff
@@ -111,15 +111,18 @@ class PresentationExchange(BaseModel):
...
+        Literal[
+            "proposal-sent",
+            "proposal-received",
+            "request-sent",
+            "request-received",
+            "presentation-sent",
+            "presentation-received",
+            "done",
+            "abandoned",
+        ],
```
TimoGlastra, line 125: State is never None I think?

TimoGlastra (changes requested):

app/generic/verifier/models.py
```diff
@@ -26,11 +26,13 @@ class SendProofRequest(ProofRequestBase):
 class CreateProofRequest(ProofRequestBase):
     proof_request: Optional[IndyProofRequest] = None
     comment: Optional[str] = None
+    connection_id: Optional[str] = None
```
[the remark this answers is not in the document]
morrieinmaas: let me double check that  - I think I added that because I got an error missing connection id. But i'll double check

app/generic/verifier/verifier_utils.py
```diff
@@ -0,0 +1,171 @@
...
+        aries_controller=aries_controller,
+        connection_id=connection_id,
+    )
+    # TODO: (In other PR) handle case where no conneciton id exists
+    # instead of simply rejecting the request
+    if not connection_record.connection_id:
+        raise CloudApiException(f"Cannot proceed. No connection ID", 404)
+
+    # Case 1: connection NOT made with publid DID
+    if connection_record.their_public_did and connection_record.their_public_did != "":
```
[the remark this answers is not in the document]
morrieinmaas: I'm quite sure there will be (or was at least) when adding this a while ago. I'll try remove it and see what happens. You're right it's better without

app/generic/verifier/verifier_utils.py
```diff
@@ -0,0 +1,171 @@
...
+    schema_ids = [rec.schema_id for rec in cred_records.results]
+    return schema_ids
+
+
+async def get_connection_from_proof(
+    aries_controller: AcaPyClient, prover: Verifier, proof_id: str
+) -> str:
+    proof_records = await prover.get_proof_records(aries_controller)
+    associated_proofs = [p for p in proof_records if p.proof_id == proof_id]
+    return associated_proofs[0].connection_id
```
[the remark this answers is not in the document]
morrieinmaas: my bad. Totally right

app/generic/verifier/verifier_utils.py
morrieinmaas: alright will change that

app/tests/e2e/test_verifier.py
morrieinmaas: good idea actually. I'll double check but I'm quite sure you're correct and there's no test with the verifier not having a public did

app/tests/util/member_personas.py
morrieinmaas: oops sorry yeah left-over from the dark ages

shared_models/shared_models.py
morrieinmaas: Added this recently. Much to my surprise it can be None/was None. I think that can happen when errors occur in aca-py. I'm not entirely sure how/why this occurs/occured but I'm sure it did occur. The annoying thing when leaving this out was that also the webhooks OpenApi stopped working when it occured because obviously the models check throws a validation error.

morrieinmaas:
Thanks for all the feedback @TimoGlastra. Will submit re-work asap

app/generic/verifier/verifier_utils.py
morrieinmaas: actually upon second thought. maybe I'm still having this somewhat wrong in my mental model of this, but what I've done is get the schemas for the request and then check whether they are all in the trust registry (the lines of code below)

```python
if await is_valid_schemas(schema_ids=schema_ids):
```

Not sure I'm missing sth or potentially making a mistake in my thinking? Thought that should cover your remarks?!

app/tests/e2e/test_verifier.py
morrieinmaas: this still needs doing (note to self @morrieinmaas )

morrieinmaas:
Another note to self and possibly question @TimoGlastra (in the accept proof request the flow seems different slightly or in other words using the same flow as in v1 yields a different webhook state than in v10. Not quite sure what is going on there. Probably don't know what I'm doing / missing something. need to discuss whether this is actually correct (and complete) or whether that should be adjusted (and/or extended)

TimoGlastra:
BTW -- @morrieinmaas I'm currently updating this PR with changes from the development branch and fixing a few things.

app/tests/e2e/test_verifier.py
TimoGlastra: I added a test for this in test_trust_registry_integration (it does a full e2e test of registering tenants, creating connections, creating schema, creating credential definition, issuing credential and finally requesting proof).

TimoGlastra:
OK -- this is now ready for review @morrieinmaas!

morrieinmaas:
Awesome! @TimoGlastra I'll give it a look/review today

app/tests/e2e/conftest.py
```diff
@@ -0,0 +1,21 @@
+import asyncio
+from typing import Any
+from aries_cloudcontroller import AcaPyClient
+import pytest
+
+from app.tests.e2e.test_fixtures import issue_credential_to_alice
+from app.tests.util.ledger import create_public_did, has_public_did
+
+
+@pytest.yield_fixture(autouse=True, scope="module")
```
morrieinmaas, line 10: nice!

app/tests/verifier/test_verifier.py
```diff
@@ -35,208 +41,281 @@
...
     assert result is presentation_exchange_record_1
     verify(VerifierV1).accept_proof_request(...)
 
+
+@pytest.mark.asyncio
+async def test_accept_proof_request_v2(mock_agent_controller: AcaPyClient):
     # V2
     when(VerifierV2).accept_proof_request(...).thenReturn(
-        get(presentation_exchange_record_2)
+        get(presentation_exchange_record_1)
```
morrieinmaas, line 173: Why is this v1 here? SHould this be v2?

app/tests/verifier/test_verifier.py
```diff
@@ -35,208 +41,281 @@
...
-            proof_id="1234",
-            presentation_spec=indy_pres_spec,
-            protocol_version="v2",
-        ),
+        presentation=presentation,
         aries_controller=mock_agent_controller,
     )
 
-    assert result is presentation_exchange_record_2
+    assert result is presentation_exchange_record_1
```
morrieinmaas, line 191: same here isn't this v2?

morrieinmaas (commented):
Great job @TimoGlastra. left two questions

app/tests/verifier/test_verifier.py
TimoGlastra: Yes, thanks!

PR #210, opened 2022-03-13
feat: no auto accept params

### LINKED ISSUE #129: how to handle testing when auto accept is enabled in acapy?

morrieinmaas:
closes #129 

disable preserve change records

all params set as wanted all test passing


TimoGlastra:
I'm not sure we should remove the exchange records for both agents. The way the current credentials API is built is that it depends on the record to be present for the holder. It is probably best to disable it for both, but we should make a separate credentials api endpoint that retrieves the actual indy credential from storage.

morrieinmaas:
alright. can you open a ticket for this, please?

PR #213, opened 2022-03-29
feat: add docs

morrieinmaas:

@TimoGlastra Any feedback welcome. 

* Anything you think should be mentioned or explained more explicitly/in more detail?
* Anything missing altogether? 
* Any changes to structuring (eg /docs folder not ok or files not ok)?
* Any other feedback.


docs/intro.md
```diff
@@ -0,0 +1,58 @@
...
+
+The standard setup spins up the the entire infrastructure required with the following containers having Swagger interfaces:
+
+- [Aca-Py Admin](http://localhost:8000/docs)
+- [Aca-Py Multitenant](http://localhost:8000/docs)
+- [Webhooks](http://localhost:3010/docs)
+- [Trust registry](http://localhost:8001/docs)
+
+In order to change specific varibales and environment parameters for each container internally you likely want to do that in the respective `.env` files. Those are localted in subfolders named correspondingly to the containers within the `environments` folder. For example,
+if you want to set `auto-accept invites` for the aca-py multitennat instance form its default value `false` to `true` you can do so by changing `ACAPY_AUTO_ACCEPT_INVITES` to `true` within `environments/yoma-multitenant/aca-py-agent.default.env`. Corollary, you can do so for other vars and other containers. Stop the container(s) accordingly and start them back up using docker-compose.
```
TimoGlastra, line 56: I'm not sure we should introduce the option to tweak e.g. the auto accept variables. The stack is currently build on auto acceptance (which we pass when creating /receiving an invitation) so auto acceptance will always be true for this.

We should probably make a distinction which environment variables are meant to be configured, and which aren't. For those that are configurable we should add some documentation, and for the other we should make it very clear in the env files they are not configurable

docs/multitenant.md
```diff
@@ -0,0 +1,7 @@
+### Multitennt agent
+
+Both aca-py agents are exposed via the same Swagger UI under the same url. They are ditinguished under the hood via the authorization method which also specifies a role. Please, rrefer to the workflows doc for further information.
+
+This agent is meant for subwallet and tenant menangement from a tenant perspective.
```
TimoGlastra, line 5: TIP: If using VSCode I can recommend the spell checker plugin to prevent typos: https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker#:~:text=Open%20up%20VS%20Code%20and,and%20reload%20window%20to%20enable.

docs/webhooks.md
```diff
@@ -0,0 +1,34 @@
+### Webhooks
```
TimoGlastra, line 1: I think it's important to mention here the webhooks websocket should not be exposed to the public web for security reasons. We don't have a websocket endpoint with authentication enabled right?

docs/workflows.md
```diff
@@ -0,0 +1,63 @@
...
+
+#### Trust registry
+
+As a client you can retrieve the trust registry (and NOT alter it). That is intential and all a consumer/client should and is able to do. Altering the trust registry is baked into admin actions and only possible with admin role.
+
+#### Webhooks
+
+A client can subscribe to the webhooks via the cloudapi (as opposed to directly via the webhooks container). This requires only targetting the `/webhooks` endpoint and optionally targetting the `/webhooks/{topic}` subroute by suffixing a topic. Using the auth mechanism (see section below), the app automatically extracts the required info about the wallet (i.e the wallet id and JWT) and retrieves only the associated webhooks with a particular wallet. Failing to authenticate will return a 403 HTTP Error.
```
TimoGlastra, line 17: How do users keep track of which webhooks they have already processed? We should mention only the last 100 webhook events can be retrieved from the api

docs/workflows.md
```diff
@@ -0,0 +1,63 @@
...
+
+### Authentication
+
+Authentication is handled by the CloudAPI and, fom a client perspective, kept simple and convenient. Either, via the Swagger UI auth (padlock butten in UI) or via the HEADER data of your client specifying an `x-api-key`. Regardless of whether ou use the UI or another client, the `x-api-key` value consists of two parts, separated by a dot:
+
+`{role}.{key/password}`
```
TimoGlastra, line 23: I'd say key/token instead of password (a password is a self-chosen value, while this is more an api key or token)

docs/workflows.md
```diff
@@ -0,0 +1,63 @@
...
+
+So, your header has the format `'x-api-key: {role}.{key/password}` which yields, for example, `'x-api-key: yoma.adminApiKey'`
+
+The first part `role` specifies the role on the surface and targets the correct endpoints under the hood and authentication mechanisms under the hood. The CloudAPI knows how to interpret the roles and will produce the correct target URLs for eg aca-py (`member` targets the multitenant agent) with the correct header expected by aca-py agent. For instance, `member` results in a `Bearer {TOKEN}` header against the multitenant agent whereas `member-admin` as role results in an `x-api-key` header for the multitenant agent (hence targeting the admin interface of the same multitennat agent). You may have noticed now that this mechanism also chooses which aca-py instance to target without having to know or specify the URL the agent resides under.
+
+Currently there are five options for `role`:
+
+- yoma
```
TimoGlastra, line 31: We should explain what each role can do. E.g. yoma is the endorser, can create schemas etc.... ecosystem-admin/member-admin can only create new tenants. ecosystem can be issuer/verifier and will be automatically registered in the trust registry, while members can only be holders

docs/workflows.md
```diff
@@ -0,0 +1,63 @@
...
+- ecosystem
+- ecosystem-admin
+- member
+- member-admin
+
+the `yoma` and `-admin` suffixed roles are admin roles. The rest are non-admin roles meaning non-admin roles have no exposure to the aca-py admin tasks nor any documented endpoints prefixed `admin:` in the cloudapi.
+
+For admin roles pass the admin password as the second part of `{role}.{key/password}`. For member/ecosystem (non-admin roles) pass the wallets JWT as the second part of `{role}.{key/password}`.
+
+### Creating schemas
```
TimoGlastra, line 41: I like that you're going into detail on how these tasks work. What I'm missing is a bit of guidance on how to build an application on this stack. What do you need to build yourself, how do you manage listening to webhooks? What are the steps involved to issue a credential and what do you need to do be able to issue a credential?

TimoGlastra (commented):
This is already looking great @morrieinmaas. What I'm missing is some practical guidance on how to use this to start building an application. There's a lot of different components, and if you're not familiar with the stack it can be really complex on how all different components fit together.

morrieinmaas:
Thanks for all the (quick0 feedback. Especially the spell checker plugin (^^ - was about time). 

I'll have a go at resolving it now

docs/workflows.md
morrieinmaas: thanks. 100% agree - token is what goes here but didn't come to mind

docs/workflows.md
morrieinmaas: Ah I removed that limit after we discussed that. Reason being that before one could receive all webhooks for all wallets and all topics which broke the swagger UI. this is  not possible any more so I removed the limit (also the main use case is via code/a client and not the swagger really)

docs/webhooks.md
morrieinmaas: good point. I mention that this shouldnt be exposed to the ww at all in the workflows doc but you're right. good to be explicit and mention it here

docs/intro.md
morrieinmaas: Fair point! To be honest I was a little more short-sighted when writing this bit and just picked any value (not caring about potentially breaking anything). That was based on the premise that we provide the settings as is (working) and showing where and how things can/should be changed (eg as opposed to in the code). If that breaks things I thought let it break and dev X can open issue Y for it. 

Your approach obviously makes sense also will change to that if desired

morrieinmaas:
@TimoGlastra Updated as discussed last week. Feel free to leave comments if anything is still missing or you want changes

morrieinmaas:
@lohanspies any feedback welcome

PR #214, opened 2022-03-30
feat: credential definition endorsement

TimoGlastra:
Add credential definition endorsement to the cloud api.

- Adds a definitions category that allows to write schemas / credential definitions.
- YOMA will write directly to the ledger, other agents will use YOMA as a transaction endorser for credential definitions
- add `wait_for_event` webhook listener that allows to wait for an event to be fired on the pubsub endpoint (means we don't have to poll for the event to happen)
- Adds JWT validation by adding a JWT secret
- Fixe some issues with the pubsub webhook event listener
- Only emit webhook for topics we support, this way we have a bit more control


TimoGlastra:
BTW @morrieinmaas not sure if this fixes the state sometimes being null for proofs

TimoGlastra:
@morrieinmaas I've increased the check_webhook_state timeout to 60s now to make tests pass, let me know if you find out why this is needed!

morrieinmaas:
> BTW @morrieinmaas not sure if this fixes the state sometimes being null for proofs

I'll see if I can still get the state null/None - 🤞 not

shared_models/shared_models.py
```diff
@@ -287,9 +287,7 @@ def v1_state_to_rfc_state(state: Optional[str]) -> Optional[str]:
         "request_received": "request-received",
         "credential_issued": "credential-issued",
         "credential_received": "credential-received",
-        # V1 uses credential_acked, while v2 uses done
-        "credential_acked": "done",
-        "credential_revoked": "credential-revoked",
+        "credential_acked": "credential-acked",
```
TimoGlastra: I think we should map the states to the same values for both v1 and v2. Was this causing issues for the tests? Now the done state for v1 is `credential-acked` while for v2 it's `done`.

shared_models/shared_models.py
```diff
@@ -133,7 +131,9 @@ class CredentialExchange(BaseModel):
         "request-received",
         "credential-issued",
         "credential-received",
+        "credential-acked",
         "done",
+        "credential-acked",
```
TimoGlastra: listed twice

morrieinmaas: will fix it up in a second

app/admin/tenants/onboarding.py
```diff
@@ -101,7 +108,11 @@ async def onboard_ecosystem_tenant(
 
 
 async def onboard_issuer(
-    *, name: str, issuer_controller: AcaPyClient, endorser_controller: AcaPyClient
+    *,
```
morrieinmaas, line 111: I know this is not a change but looking at this rn I'm wondering why we have the * unpacking here?

app/dependencies.py
```diff
@@ -15,29 +18,58 @@
...
+    else:
+        try:
+            # Decode JWT
+            token_body = jwt.decode(
+                auth.token, ACAPY_MULTITENANT_JWT_SECRET, algorithms=["HS256"]
+            )
+        except jwt.InvalidTokenError:
+            raise HTTPException(403, "Unauthorized")
+
+        wallet_id = token_body.get("wallet_id")
```
morrieinmaas, line 65: why do we use .get here?

app/facades/acapy_ledger.py
```diff
@@ -159,3 +159,37 @@ async def write_credential_def(controller: AcaPyClient, schema_id: str) -> str:
...
+            from which to identify a schema
+
+    Returns:
+    -------
+    schema_id : string
+    """
+    # scrape schema id or sequence number from cred def id
+    tokens = credential_definition_id.split(":")
+    if len(tokens) == 8:  # node protocol >= 1.4: cred def id has 5 or 8 tokens
+        return ":".join(tokens[3:7])  # schema id spans 0-based positions 3-6
```
morrieinmaas, line 185: so in this case the schema id is baked in the cred def id?!

app/generic/definitions.py
```diff
@@ -0,0 +1,390 @@
...
+    get_credential_definition_futures = [
+        aries_controller.credential_definition.get_cred_def(
+            cred_def_id=credential_definition_id
+        )
+        for credential_definition_id in credential_definition_ids
+    ]
+
+    # Wait for completion of retrieval and transform all credential definitions
+    # into response model (if a credential definition was returned)
+    credential_definition_results = await asyncio.gather(
```
morrieinmaas, line 125: I think you can just put await in the list comprehension above

morrieinmaas:
```python
credential_definition_results = [
        await aries_controller.credential_definition.get_cred_def(
            cred_def_id=credential_definition_id
        )
        for credential_definition_id in credential_definition_ids
    ]
```

app/generic/definitions.py
```diff
@@ -0,0 +1,390 @@
...
+        *get_credential_definition_futures
+    )
+    credential_definitions = [
+        _credential_definition_from_acapy(credential_definition.credential_definition)
```
morrieinmaas, line 129: Think you could also do 
```python
credential_definitions = [
   CredentialDefinition(**credential_definition.credential_definition)
   for credential_definition in credential_definition_results
   if credential_definition.credential_definition
]
```

app/generic/definitions.py
```diff
@@ -0,0 +1,390 @@
...
+    return CredentialSchema(
+        id=schema.id,
+        name=schema.name,
+        version=schema.version,
+        attribute_names=schema.attr_names,
+    )
+
+
+def _credential_definition_from_acapy(credential_definition: AcaPyCredentialDefinition):
+    return CredentialDefinition(
```
morrieinmaas, line 72: or as below comment (maybe good to keep this method as you're reusing it
```python
def _credential_definition_from_acapy(credential_definition: AcaPyCredentialDefinition):
    return CredentialDefinition(**credential_definition)
```

app/generic/definitions.py
```diff
@@ -0,0 +1,390 @@
...
+            sig_type = signature["operation"]["signature_type"]
+            schema_ref = signature["operation"]["ref"]
+            tag = signature["operation"]["tag"]
+            credential_definition_id = f"{public_did}:3:{sig_type}:{schema_ref}:{tag}"
+        except Exception as e:
+            raise CloudApiException(
+                "Unable to construct credential definition id from signature response"
+            ) from e
+
+        # FIXME: ACA-Py 0.7.3 has no way to associate the credential definition id with a transaction record
```
morrieinmaas, line 254: Are we leaving this out for now then?

app/generic/definitions.py
```diff
@@ -0,0 +1,390 @@
...
+
+    # Initiate retrieving all schemas
+    schema_ids = response.schema_ids or []
+    get_schema_futures = [
+        aries_controller.schema.get_schema(schema_id=schema_id)
+        for schema_id in schema_ids
+    ]
+
+    # Wait for completion of retrieval and transform all schemas into response model (if a schema was returned)
+    schema_results = await asyncio.gather(*get_schema_futures)
```
morrieinmaas, line 319: Same as above - I think you can just do await in the list comprehension?

app/main.py
```diff
@@ -37,12 +37,21 @@
...
 app.include_router(tenants.router)
 app.include_router(verifier.router)
-app.include_router(schemas.router)
-app.include_router(credential_definitions.router)
 app.include_router(trust_registry.router)
+app.include_router(definitions.router)
 app.include_router(webhooks.router)
 
 
+@app.on_event("shutdown")
```
morrieinmaas, line 45: Ah nice!

app/tests/e2e/test_definitions.py
```diff
@@ -0,0 +1,180 @@
...
+
+    # when
+    result = (
+        await definitions.create_credential_definition(
+            credential_definition, yoma_acapy_client, auth
+        )
+    ).dict()
+
+    assert_that(result).has_tag(credential_definition.tag)
+    # FIXME: schema_is is seq_no. Should update to schema id instead
```
morrieinmaas, line 63: schema_id instead of schema_is?

app/webhook_listener.py
```diff
@@ -0,0 +1,115 @@
...
+        else:
+            Webhooks.off(on_webhook)
+
+    async def stop_listener():
+        Webhooks.off(on_webhook)
+
+    # Subscribe to webhook events and put them in a queue
+    await Webhooks.on(on_webhook)
+
+    return wait_for_event_with_timeout, stop_listener
```
morrieinmaas, line 115: nice

shared_models/__init__.py
```diff
@@ -1,15 +1,20 @@
...
     IssueCredentialProtocolVersion,
     presentation_record_to_model,
     pres_id_no_version,
+    RedisItem,
+    BasicMessage,
+    PayloadType,
+    topic_mapping,
+    WEBHOOK_TOPIC_ALL,
+    CloudApiTopics,
+    Endorsement,
```
morrieinmaas, line 19: nice

webhooks/main.py
```diff
@@ -50,60 +49,65 @@ async def wallet_root(
...
     request: Request,
     service: Service = Depends(Provide[Container.service]),
 ):
-    payload = await request.json()
+
     try:
-        wallet_id = {"wallet_id": request.headers["x-wallet-id"]}
+        wallet_id = request.headers["x-wallet-id"]
     except KeyError:
-        wallet_id = {"wallet_id": "admin"}
```
morrieinmaas, line 71: nice

morrieinmaas (approved):
Great effort @TimoGlastra. Left a few questions and suggestions

app/generic/definitions.py
TimoGlastra: I think this will make the retrieving of credential definitions sequential where asyncio.gather will make it run in parallel. This can have a significant impact on time it will take to finish processing

app/generic/definitions.py
TimoGlastra: I'm not really a fan of this implicit approach tbh. We're transforming properties between two different data models, and although the properties currently match they may drift apart over time. It also explains to the developer how we're transforming from the acapy model to our own model. So I mainly do it to be explicit.

app/generic/definitions.py
TimoGlastra: Ah no I found a workaround for this (see `credential_definition_id = f"{public_did}:3:{sig_type}:{schema_ref}:{tag}"` above). Will remove 👍

app/generic/definitions.py
TimoGlastra: Same, want to run in parallel

app/dependencies.py
TimoGlastra: To not error out when the property doesn't exist. .get() returns None by default so we can handle the wallet not exists a few lines below.

Could change it to a try/except but feels quite verbose to me and still doesn't capture the property exists but value is None (as that won't throw an eror)

app/facades/acapy_ledger.py
TimoGlastra: Yes

app/admin/tenants/onboarding.py
morrieinmaas: thanks for clarifying

PR #224, opened 2022-04-28
Add exception handling to webhooks listener.

### LINKED ISSUE #215: Change the way the webhooks validates models re: unknown webhook event.
Change the way web hooks validates models. Currently it is causing an error. If it doesn't handle the webhook right, or if it doesnt recieve an webhook from AcaPy rather than an error - it should not panic. I t should be able to know that it's received "something."

morrieinmaas:
If anything goes wrong in creating the return data for the webhook
just skip it and log it.

Quick and simple addition. 

Closes #215 


TimoGlastra (approved):
Nice, LGTM 👍

PR #231, opened 2022-05-11
refactor: combine member and ecosystem roles

TimoGlastra:
Combines the member and ecosystem roles into a single tenant role. Permissions are determined by the trust registry rules. 

Nest step is extracting the trust registry  / yoma specific code into a separate module. I'd like to have a discussion about this friday as there's a few approaches we could take here with different implications.

### DIFF
--- a/app/admin/tenants/tenants.py
+++ b/app/admin/tenants/tenants.py
@@ -10,7 +10,6 @@ from aries_cloudcontroller import (
     RemoveWalletRequest,
     UpdateWalletRequest,
 )
-from aries_cloudcontroller.model.wallet_record import WalletRecord
 from fastapi import APIRouter, Depends
 
 from app.admin.tenants.models import (
@@ -22,8 +21,8 @@ from app.admin.tenants.models import (
     tenant_from_wallet_record,
 )
 from app.admin.tenants.onboarding import (
-    handle_ecosystem_tenant_update,
-    onboard_ecosystem_tenant,
+    handle_tenant_update,
+    onboard_tenant,
 )
 from app.dependencies import AcaPyAuth, Role, acapy_auth, agent_role
 from app.error import CloudApiException
@@ -33,13 +32,12 @@ from app.facades.trust_registry import (
     register_actor,
     remove_actor_by_id,
 )
-from app.role import AgentType
 
 logger = logging.getLogger(__name__)
 
 router = APIRouter(prefix="/admin/tenants", tags=["admin: tenants"])
 
-multitenant_admin = agent_role([Role.MEMBER_ADMIN, Role.ECOSYSTEM_ADMIN])
+multitenant_admin = agent_role(Role.TENANT_ADMIN)
 
 
 def tenant_api_key(role: Role, tenant_token: str):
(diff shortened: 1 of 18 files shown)


app/tests/test_dependencies.py
```diff
@@ -245,32 +203,19 @@ async def make_call(route_suffix: str = "", headers=None):
...
-    assert injected_controller.base_url == Role.MEMBER.agent_type.base_url
-    assert injected_controller.client.headers["x-api-key"] == "provided-x-api-key-1"
-    assert isinstance(injected_controller, AcaPyClient)
-
 
 @pytest.mark.asyncio
-async def test_ecosystem_admin_agent():
-    async with asynccontextmanager(dependencies.agent_role(role=Role.ECOSYSTEM_ADMIN))(
-        auth=AcaPyAuth(role=Role.ECOSYSTEM_ADMIN, token=GOVERNANCE_ACAPY_API_KEY)
+async def test_tenant_admin_agent():
```
morrieinmaas, line 216: This test/a test with this name already exists in line 43. Looks duplicate to me. Remove one if so or change the logic to what you wanna test

morrieinmaas (changes requested):
Overall LGTM🎉🙌. Think there is a duplicate test in `test_dependencies` though

app/admin/tenants/tenants.py
```diff
@@ -123,13 +69,13 @@ async def create_tenant(
             key_management_mode="managed",
             label=body.name,
             wallet_key=token_urlsafe(48),
-            wallet_name=get_wallet_name_for_role(tenant_role),
+            wallet_name=uuid4().hex,
```
morrieinmaas, line 72: Is this really what we want here? not a human readable name?

app/tests/e2e/test_tenants.py
```diff
@@ -200,17 +201,17 @@ async def test_create_tenant_ecosystem_verifier(
     assert_that(tenant).has_tenant_name(name)
     assert_that(tenant).has_created_at(wallet.created_at)
     assert_that(tenant).has_updated_at(wallet.updated_at)
-    assert wallet.settings["wallet.name"].startswith("ecosystem.")
+    assert_that(wallet.settings["wallet.name"]).is_type_of(str)
```
morrieinmaas, line 204: Maybe (if I'm getting this correctly from above) this is uuid.hex so a string of certain format/lnegth. In which case we can check specifically for that instead of just string?!

app/tests/e2e/test_tenants.py
```diff
@@ -251,13 +252,13 @@ async def test_update_tenant_ecosystem_verifier_to_issuer(
...
```
morrieinmaas, line 255: same as above - possibly specific check for length and/or format instead of type

app/tests/e2e/test_tenants.py
```diff
@@ -311,15 +312,13 @@ async def test_update_tenant_ecosystem_verifier_to_issuer(
     assert_that(new_tenant).has_image_url(new_image_url)
     assert_that(new_tenant).has_tenant_name(new_name)
     assert_that(new_tenant).has_created_at(wallet.created_at)
-    assert wallet.settings["wallet.name"].startswith("ecosystem.")
+    assert_that(wallet.settings["wallet.name"]).is_type_of(str)
```
morrieinmaas, line 315: same as above

app/tests/test_dependencies.py
morrieinmaas: have you removed this duplicate?

morrieinmaas (commented):
Few nit-picks/questions - but LGTM!🙌

app/admin/tenants/tenants.py
TimoGlastra: The wallet name is something internal and more of an id to the wallet. If you look at the indy-sdk docs you can see it uses id instead of name: https://github.com/hyperledger/indy-sdk/tree/master/wrappers/nodejs#createwallet--config-credentials----void. Not sure why ACA-Py chose to use name here. A name suggests to me there can be multiple wallets with the same name, but that's not the case, so it acts more of an id here

morrieinmaas: ah alright thx for the pointer i'll read up on that

app/tests/e2e/test_tenants.py
TimoGlastra: Can I address these in a follow up PR? I have a PR that is dependant on these changes so to avoid a lot of hassle it would be easier to address in that pr (and if so, could you approve)

TimoGlastra: This has been addressed in #232

app/tests/e2e/test_tenants.py
TimoGlastra: This has been addressed in #232

app/tests/e2e/test_tenants.py
TimoGlastra: This has been addressed in #232

morrieinmaas (approved):

PR #233, opened 2022-05-30
feat: use askar wallet

TimoGlastra:
Use askar wallet (and indy-vdr, indy-credx shared components) instead of the indy wallet


morrieinmaas (approved):
LGTM - nice and tidy

PR #240, opened 2022-06-07
added dos2unix hack to agent dockerfile for windows builds

jasondicker:
On a Windows machine, I received the following error on container startup:

/bin/sh: 1: ./startup.sh: not found

To resolve this, I added a 'dos2unix' step in the agent dockerfile. Please confirm working on linux machines.


morrieinmaas:
If I'm not mistaken dos2unix handles line endings(which differ on Windows and Unix). maybe it's better to just fix the line endings or run this from bash in WSL instead?

jasondicker:
Thanks, I'll try again without requiring a change to the repo.

morrieinmaas:
Thanks for giving it another go

PR #251, opened 2022-09-04
bump version aca-py

### LINKED ISSUE #249: Update to 0.7.4
- Update Cloud controller to 0.7.4
- Update cloud api to work with changes from 0.7.4

morrieinmaas:
closes #249 


morrieinmaas:
error related to [this issue](https://github.com/minio/minio/issues/8161) (weirdly enough as that should be fixed

morrieinmaas:
fails by creating pub did with endorsement (fixing that - see linked issue above - should fix most failing tests)

app/admin/tenants/onboarding.py
```diff
@@ -126,94 +124,129 @@ async def onboard_issuer(
...
+        )
+
+        # Wait for connection to be completed before continuing
+        try:
+            await wait_for_event(
+                filter_map={
+                    "connection_id": connection_record.connection_id,
+                    "state": "completed",
+                }
+            )
```
TimoGlastra, line 181: This is a bit useless I think if we have the completed listener for the endorser, and we aren't using the event data. I think we can just remove the connection completed listener for the new tenant completely?

app/admin/tenants/onboarding.py
```diff
@@ -126,94 +124,129 @@ async def onboard_issuer(
...
-                "connection_id": connection_record.connection_id,
-                "state": "completed",
-            }
+        # TODO: This doesn't need endorsement as of 0.7.5-rc0 onward - bug in 0.7.4
+        # Change this in future versions
+        await acapy_wallet.set_public_did(
+            issuer_controller,
+            did=issuer_did.did,
+            connection_id=connection_record.connection_id,
+            create_transaction_for_endorser=True,
```
TimoGlastra, line 226: I think we can just set this to false and it will work without endorsement (because attrib transactions are free). This means we don't have to endorse the transaction, meaning it will be simpler and quicker

app/admin/tenants/onboarding.py
```diff
@@ -126,94 +124,129 @@ async def onboard_issuer(
...
+        try:
+            await endorser_wait_for_transaction(
+                filter_map={
+                    "state": "request-received",
+                }
+            )
+        except TimeoutError:
+            raise CloudApiException("Error creating connection with endorser", 500)
+
+        txns = await endorser_controller.endorse_transaction.get_records()
```
TimoGlastra, line 239: As per my previous comment, I don't think we need to endorse the transaction. However if we do need it, I think we could take the transaction_id from the `endorser_wait_for_transaction` return value event, and that would save us to retrieve _all_ endorsement requests (which wouldn't really scale)

app/facades/acapy_wallet.py
```diff
@@ -60,7 +60,13 @@ async def create_did(controller: AcaPyClient) -> Did:
     return Did(did=did_result.result.did, verkey=did_result.result.verkey)
 
 
-async def set_public_did(controller: AcaPyClient, did: str) -> DID:
+async def set_public_did(
+    controller: AcaPyClient,
+    did: str,
+    connection_id: str = None,
+    create_transaction_for_endorser: bool = False,
+    is_endorsed: bool = False,
```
TimoGlastra, line 68: This feels a bit misleading. What happens if is_endorsed is True and there is no result.result? there won't be a did returned right (which the return type of this method says it does). Maybe we can add an union return type that also takes into account the endorsement return type. This way we don't have to pass is_endorsed which sometimes can hard to know

app/generic/connections/connections.py
```diff
@@ -137,7 +139,8 @@ async def connect_to_public_did(
         their_public_did=body.public_did
     )
 
-    return conn_record_to_connection(conn_record)
+    # return conn_record_to_connection(conn_record)
+    return conn_record
```
TimoGlastra, line 143: Why this change? We should return our connection model, not the AcaPy ConnRecord

app/generic/connections/connections.py
```diff
@@ -172,18 +175,18 @@ async def accept_invitation(
...
     )
-    return conn_record_to_connection(conn_record)
+    return conn_record_to_connection(connection_record)
 
 
 @router.get("/", response_model=List[Connection])
 async def get_connections(
     aries_controller: AcaPyClient = Depends(agent_selector),
-):
+) -> List[Optional[Connection]]:
```
TimoGlastra, line 189: I'm not sure this is the correct type. List[Optional[Connection]] means there could be None values in the list ([None, Conn, None, None]), but that's not the case right (and otherwise _shouldn't_ be the case).

There is the possiblity of returning an empty list, but in that case the type `List[Connection]` still is correct

app/generic/connections/connections.py
```diff
@@ -158,11 +161,11 @@ async def create_invitation(
     return invitation
 
 
-@router.post("/accept-invitation", response_model=Connection)
+@router.post("/accept-invitation")
```
TimoGlastra, line 164: Does the openapi definition still infer the return type correctly if we don't pass response_model?

app/generic/definitions.py
```diff
@@ -353,12 +353,23 @@ async def create_schema(
...
+            body=schema_send_request, create_transaction_for_endorser=False
+        )
+    except ClientResponseError as e:
+        if e.status == 400 and "already exist" in e.message:
+            schema_ids = await aries_controller.schema.get_created_schemas(
+                schema_name=schema.name
+            )
+            schema = await aries_controller.schema.get_schema(
+                schema_id=schema_ids.schema_ids[0]
+            )
```
TimoGlastra, line 367: There could be multiple schema's with the same name. If the schema already exists we can construct the schema id based on the did, schema_name and schema_version. Otherwise we risk returning a different schema that is unrelated to the schema we were trying to create.

One thing also to note is that the attributes we passed to create the schema could different from the attributes that are present in the existing schema. We should probably do a check for this, or just throw an error like before that the schema already exists

app/tests/e2e/test_messaging.py
```diff
@@ -17,6 +18,8 @@ async def test_send_trust_ping(
         connection_id=bob_and_alice_connection["alice_connection_id"], comment="Donda"
     )
 
+    # Wait for connection to be ready - sometimes this takes a second or two
+    sleep(5)
```
TimoGlastra, line 22: Maybe we can add an event listener to the bob_and_alice_connection so we don't have to add timeouts throughout the codebase? That way it's fixed once for all tests that use `bob_and_alice_connection`

app/tests/e2e/test_tenants.py
```diff
@@ -271,38 +271,9 @@ async def test_update_tenant_verifier_to_issuer(
...
-            client,
-            "connections",
-            {
-                "state": "completed",
-                "connection_id": endorser_connection.connection_id,
-            },
-        )
-
-    # Connection invitation
-    assert_that(endorser_connection).has_their_public_did(endorser_did.did)
```
TimoGlastra, line 301: Why did you remove this part of the test? It checks whether the endorser connection uses the public did to connect

app/tests/admin/test_onboarding.py
```diff
@@ -70,27 +72,12 @@ async def test_onboard_issuer_public_did_exists(
         )
     )
     onboard_result = await onboarding.onboard_issuer(
-        name="issuer_name",
```
TimoGlastra, line 73: Why did you remove the name? This is used in the trust registry I think

app/tests/util/member_personas.py
```diff
@@ -80,6 +83,69 @@ async def bob_and_alice_public_did(
     alice_acapy_client: AcaPyClient,
     bob_acapy_client: AcaPyClient,
 ) -> BobAlicePublicDid:
+    tenant_admin_acapy = governance_acapy_client()
```
TimoGlastra, line 86: tenant admin is the admin user of the tenant agent, while this is the governance agent. To avoid confusion maybe we can call it governance_admin_acapy?

```suggestion
    governance_admin_acapy = governance_acapy_client()
```

app/tests/util/member_personas.py
```diff
@@ -80,6 +83,69 @@ async def bob_and_alice_public_did(
...
+        await alice_acapy_client.endorse_transaction.set_endorser_role(
+            conn_id=alice_conn.connection_id, transaction_my_job="TRANSACTION_AUTHOR"
+        )
+    )
+    set_endorser_info_result_alice = (
+        await alice_acapy_client.endorse_transaction.set_endorser_info(
+            conn_id=alice_conn.connection_id,
+            endorser_did=set_public_did_result.did,
+            endorser_name="endorser",
+        )
```
TimoGlastra, line 136: Why are you setting up endorsement between alice and bob? Is this because of the bug present in ACA-Py 0.7.4?

docker-compose.yaml
```diff
@@ -10,6 +10,7 @@ services:
     volumes:
       - ./shared_models/shared_models.py:/app/shared_models.py
       - ./app:/app
+      # - ./aries-cloudcontroller-python:/aries-cloudcontroller-python
```
TimoGlastra, line 13: Can this be removed?

docker-compose.yaml
```diff
@@ -84,6 +85,7 @@ services:
...
```
TimoGlastra, line 88: same here

dockerfiles/agents/Dockerfile.agent
```diff
@@ -1,21 +1,16 @@
...
-
+FROM bcgovimages/von-image:py36-1.16-1
 
 USER root
 RUN apt-get update
 RUN apt-get update && apt-get install -y gcc
 
-
-RUN pip3 install aries-cloudagent[askar]==0.7.3
+RUN pip3 install aries-cloudagent[askar,indy]==0.7.4
```
TimoGlastra, line 7: I think we can remove the indy requirement here

dockerfiles/agents/Dockerfile.author.agent
```diff
@@ -0,0 +1,18 @@
+FROM bcgovimages/von-image:py36-1.16-1
+
+USER root
+RUN apt-get update
+RUN apt-get update && apt-get install -y gcc
+
+RUN pip3 install aries-cloudagent[askar,indy]==0.7.4
```
TimoGlastra, line 7: Can we also use the `aries-cloudagent:py36-1.16-1_0.7.4` as base here? Would be good to have a consistent docker setup

dockerfiles/fastapi/Dockerfile
```diff
@@ -1,13 +1,16 @@
 FROM tiangolo/uvicorn-gunicorn-fastapi:python3.8
 
 COPY app /app/
+# COPY aries-cloudcontroller-python /aries-cloudcontroller-python/
```
TimoGlastra, line 4: can be removed?

junit.xml
```diff
@@ -0,0 +1 @@
+<?xml version="1.0" encoding="utf-8"?><testsuites><testsuite name="pytest" errors="0" failures="0" skipped="0" tests="9" time="54.379" timestamp="2022-10-21T15:32:39.352633" hostname="felurian"><testcase classname="tests.e2e.test_connections" name="test_create_invitation" time="17.893" /><testcase classname="tests.e2e.test_connections" name="test_accept_invitation" time="13.344" /><testcase classname="tests.e2e.test_connections" name="test_get_connections" time="0.898" /><testcase classname="tests.e2e.test_connections" name="test_get_connection_by_id" time="0.422" /><testcase classname="tests.e2e.test_connections" name="test_delete_connection" time="0.575" /><testcase classname="tests.e2e.test_connections" name="test_bob_and_alice_connect" time="2.722" /><testcase classname="tests.e2e.test_connections" name="test_create_invitation_oob" time="0.237" /><testcase classname="tests.e2e.test_connections" name="test_accept_invitation_oob" time="0.781" /><testcase classname="tests.e2e.test_connections" name="test_oob_connect_via_public_did" time="17.476" /></testsuite></testsuites>
```
TimoGlastra, line 1: can be remvoed?

shared_models/shared_models.py
```diff
@@ -100,13 +103,54 @@ class Endorsement(BaseModel):
...
+    connection_id: Optional[str] = None
+    updated_at: Optional[str] = None
+    attach_thread_id: Optional[str] = None
+    multi_use: Optional[bool] = False
+    our_recipient_key: Optional[str] = None
+    their_service: Optional[Union[ServiceDecorator, Any]] = None
+    our_service: Optional[Union[ServiceDecorator, Any]] = None
+
+
+class OobRecord(BaseModel):
```
TimoGlastra, line 130: Why do we need to type OobRecord? Is this not exported from ACA-Py? (due to the bug?)

shared_models/shared_models.py
```diff
@@ -100,13 +103,54 @@ class Endorsement(BaseModel):
...
+    nvi_msg_id: Optional[str]
```
TimoGlastra, line 131:
```suggestion
    invi_msg_id: Optional[str]
```

shared_models/shared_models.py
```diff
@@ -100,13 +103,54 @@ class Endorsement(BaseModel):
     state: str
 
 
+class ServiceDecorator(TypedDict):
+    endpoint: str
+    recipient_keys: List[str]
+    routing_keys: Optional[List[str]]
+
+
+class Oob(BaseModel):
```
TimoGlastra, line 112: It would be good to use literal/enum/union types as much as possible (e.g. for state/role etc...)

shared_models/shared_models.py
```diff
@@ -100,13 +103,54 @@ class Endorsement(BaseModel):
...
+    role: Optional[str] = str
+    invi_msg_id: Optional[str] = None
+    state: Optional[str] = str
+    trace: Optional[str] = None
+    our_recipient_key: Optional[str] = None
+    oob_id: Optional[str] = None
+    invitation: Union[InvitationMessage, Mapping[str, Any]] = None
```
TimoGlastra, line 119: I think this should be `InvitationMessage`. If the type isn't `InvitationMessage` when transforming, we should transform it to `InvitationMessage` (Otherwise we let the user deal with the incosistency instead of handling it for them)

TimoGlastra: Same for `their_service` and `our_service`.

shared_models/shared_models.py
```diff
@@ -100,13 +103,54 @@ class Endorsement(BaseModel):
...
+    created_at: Optional[str] = None
+    connection_id: Optional[str] = None
+    updated_at: Optional[str] = None
+    attach_thread_id: Optional[str] = None
+    multi_use: Optional[bool] = False
+    our_recipient_key: Optional[str] = None
+    their_service: Optional[Union[ServiceDecorator, Any]] = None
+    our_service: Optional[Union[ServiceDecorator, Any]] = None
+
```
TimoGlastra, line 128: Are these important to expose? We can always keep the model ligth for now and add properties as needed. The smaller the api / number of properties, the easier it will be for us to change it later.

TimoGlastra (changes requested):

morrieinmaas:
thx for the feedback. will address this today

shared_models/shared_models.py
morrieinmaas: Agreed. I don't think we do either. Just kept all of the ones from aca-py

shared_models/shared_models.py
morrieinmaas: thx well spotted

app/admin/tenants/onboarding.py
morrieinmaas: Let me try this again. I think your suggestion works in 0.7.5-rc0 onwards (and current main) but not in 0.7.4 - but will give it a try

app/admin/tenants/onboarding.py
morrieinmaas: see comment below

app/facades/acapy_wallet.py
morrieinmaas: If it is endorsed the set public did succeeds but no result.result is returned (in 0.7.4)

app/admin/tenants/onboarding.py
morrieinmaas: ah yeah I think you're correct. I initially thought we'd need to wit for connection completed but as you say waiting for endorser connection to be completed should suffice

app/admin/tenants/onboarding.py
morrieinmaas: If we don't create transaction for endorser and don't endorse this (in 0.7.4) we get `'Cannot connect via public DID that has no associated services.`

app/facades/acapy_wallet.py
morrieinmaas: So this prevents the throwing an error when result.result is not present

app/admin/tenants/onboarding.py
morrieinmaas: thx for the hint here. extracting it from `endorser_wait_for_transaction` is the better way

shared_models/shared_models.py
morrieinmaas: Couldnt import that from version 0.7.4

app/tests/util/member_personas.py
morrieinmaas: No this isn't needed I think. Sorry, leftover from experimenting

app/tests/admin/test_onboarding.py
morrieinmaas: also left over from debugging with different versions

morrieinmaas: will put it back

app/admin/tenants/onboarding.py
TimoGlastra: Ah okay nvm in that case

app/facades/acapy_wallet.py
TimoGlastra: Yes, but the return type is misleading in that case. Could we maybe trigger it based on the return status code instead in that case? (So if no error is thrown we see it as being succeeded). is_endorsed requires deep knowledge of the setup, which may not always be known.

morrieinmaas: ah controller doesn't return the status code if i'm not mistaken. I'll think of something else

app/tests/e2e/test_tenants.py
morrieinmaas: ah darn - forgot to add that back in. 
Now getting token invalid error for the subwallet after updating. Am on it.

Issue #261, opened 2022-10-26
Update to aca-py 0.7.5


morrieinmaas:
aca-py 0.7.5 should be released soon. This release will include some patches for endorsement among other things. we should probably upgrade to this version in the near future.

- [ ] create cloudcontroller version for 0.7.5
- [ ] update cloudapi to use 0.7.5 and corresponding cloudcontroller
- [ ] adjust onboarding flow (and other places) using endorsement
- [ ] adjust tests

TimoGlastra:
I think 0.7.5 has been released now

PR #262, opened 2022-10-26
feat: add revocation support

### LINKED ISSUE #204: Support revocation
We should look at supporting revocation in the cloud api. This is mostly already handled by ACA-Py so we should just look at what is needed in terms of infrastructure and API endpoints. 

- [ ] Add tails server to the infrastructure and provide this to all ACA-Py agents
- [ ] Allow to specify `supports_revocation` when creating a credential definition
- [ ] Add endpoint to revoke a credential by the credential exchange record id or the indy specific identifiers.
- [ ] add tests

This is not a requirement for the V1 Must Have

morrieinmaas:
closes #204


morrieinmaas:
requires merge of #261 related PR

app/generic/definitions.py
```diff
@@ -248,11 +257,69 @@ async def create_credential_definition(
...
+                            },
+                            timeout=30,
+                        )
+                    except TimeoutError:
+                        raise CloudApiException(
+                            "Failed to retrieve transaction record for endorser", 500
+                        )
+                    finally:
+                        await stop_listener()
+
```
TimoGlastra, line 311: starting the listener after the actions that triggers the event can possibly lead to race conditions

app/generic/definitions.py
```diff
@@ -248,11 +257,69 @@ async def create_credential_definition(
             raise CloudApiException(
                 "Unable to construct credential definition id from signature response"
             ) from e
-
     else:
-        await stop_listener()
         credential_definition_id = result.credential_definition_id
 
+    if credential_definition.support_revocation:
```
TimoGlastra, line 263: Does this need to be done manually? I know they autmoated 99 percent of it, and thought this was also handled, but not 100% sure

morrieinmaas: Hmm let me double check this, but I didn't get this to work otherwise

morrieinmaas:
@TimoGlastra thx for the feedback already. I still need to add some tests and e2e tests for the revoking a credential. So it's WIP. Just added it and manually verified it uses revocation registry

app/generic/definitions.py
morrieinmaas: so not doing this leaves credential issuance in 'abandoned' state. 

I'll get to the bottom of it

lohanspies:
> @morrieinmaas will take a look at this one when all other PRs are merged and this one is up to date with main

@TimoGlastra can you perhaps review and close this PR?

PR #263, opened 2022-11-01
feat: update to aca-py 0.7.5

### LINKED ISSUE #261: Update to aca-py 0.7.5

morrieinmaas:

closes #261


morrieinmaas:
still need to double check endorser flow - should now work bug free so simplification should be possible

app/admin/tenants/onboarding.py
```diff
@@ -211,15 +211,13 @@ async def onboard_issuer(
         )
 
         await acapy_ledger.accept_taa_if_required(issuer_controller)
-        # TODO: This doesn't need endorsement as of 0.7.5-rc0 onward - bug in 0.7.4
-        # Change this in future versions
+        # NOTE: Still needs endorsement in 0.7.5 release
+        # Otherwise did has no associated services.
```
TimoGlastra, line 215: What is exactly going wrong here? Thought it would be fixed in 0.7.5. If we can find the root cause we can open an issue in ACA-Py

TimoGlastra (approved):

app/admin/tenants/onboarding.py
morrieinmaas: I'm still trying to find that out. The error is something along the lines of "cannot to connect via public did with no associated services".

TimoGlastra:
Can this be merged?

morrieinmaas:
> Can this be merged?

Short answer: yes. There is still a few things with endorsement that work but not the expected way. However, Those need to be addressed in aca-py (and the controller eventually). So feel free to merge

Issue #265, opened 2022-11-09
Support connectionless proof request


TimoGlastra:
We need to disable the check for the did in the create-proof-request (not for the schema) as it prevents from creating any proof request.

The flow that should work after this:
- (Prover) call `/generic/verifier/create-request`
- (Prover) call `/generic/connections/oob/create-inivtation`
  - `create_connection` should be false
  - proof request message should be added to the invitation
- (Prover) send oob invitation out of band to holder
- (Holder) call `/generic/connections/oob/accept-invitation`
- (Holder) wait for proofs event for the received proof request
- (Holder) send presentation based on proof request

TimoGlastra:
@morrieinmaas this was requested for the yoma platform, can you take a look at this? I think it should be quite straightforward as al the steps are there, they just need to be tweaked a bit to work together

morrieinmaas:
@TimoGlastra sure thing - lemme have a look at this

PR #266, opened 2022-11-10
feat: enable connectionless proof

morrieinmaas:
* create invite works
* accept invite works
* both OOB
* BUT: no proof request is going through

### DIFF
--- a/app/generic/verifier/models.py
+++ b/app/generic/verifier/models.py
@@ -27,7 +27,6 @@ class CreateProofRequest(BaseModel):
 
 class AcceptProofRequest(BaseModel):
     proof_id: str
-    connectionless: Optional[bool] = False
     presentation_spec: IndyPresSpec
 
 
--- a/app/generic/verifier/models.py
+++ b/app/generic/verifier/models.py
@@ -9,8 +9,8 @@ from pydantic import BaseModel
 
 
 class PresentProofProtocolVersion(Enum):
-    v1 = "v1"
-    v2 = "v2"
+    v1: str = "v1"
+    v2: str = "v2"
 
 
 class SendProofRequest(BaseModel):
--- a/app/generic/verifier/models.py
+++ b/app/generic/verifier/models.py
@@ -27,6 +27,7 @@ class CreateProofRequest(BaseModel):
 
 class AcceptProofRequest(BaseModel):
     proof_id: str
+    connectionless: Optional[bool] = False
     presentation_spec: IndyPresSpec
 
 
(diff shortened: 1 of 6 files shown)


morrieinmaas:
closes #265

app/generic/verifier/verifier.py
```diff
@@ -239,9 +238,12 @@ async def accept_proof_request(
     try:
         prover = __get_verifier_by_version(presentation.proof_id)
 
-        await assert_valid_prover(
-            aries_controller=aries_controller, prover=prover, presentation=presentation
-        )
+        if not presentation.connectionless:
```
TimoGlastra, line 241: We can check whether the proof is connectionless by first retrieving it from the ACA-Py instance. This prevents the user needing to provide info that is already availalbe to use and also removes a way for inconsistent state: the user sets connectionless: false, but the exchange is actually connectionless

app/generic/verifier/models.py
```diff
@@ -27,6 +27,7 @@ class CreateProofRequest(BaseModel):
 
 class AcceptProofRequest(BaseModel):
     proof_id: str
+    connectionless: Optional[bool] = False
```
TimoGlastra, line 30:
```suggestion
```

app/generic/verifier/verifier.py
```diff
@@ -177,9 +176,9 @@ async def send_proof_request(
     try:
         prover = __get_verifier_by_version(proof_request.protocol_version)
 
-        await assert_valid_verifier(
-            aries_controller=aries_controller, proof_request=proof_request
-        )
+        # await assert_valid_verifier(
+        #     aries_controller=aries_controller, proof_request=proof_request
+        # )
```
TimoGlastra, line 181: I think we only want to disable this for connectionless proof requests

```suggestion
        await assert_valid_verifier(
            aries_controller=aries_controller, proof_request=proof_request
        )
```

app/generic/verifier/verifier.py
```diff
@@ -246,9 +238,12 @@ async def accept_proof_request(
     try:
         prover = __get_verifier_by_version(presentation.proof_id)
 
-        await assert_valid_prover(
-            aries_controller=aries_controller, prover=prover, presentation=presentation
-        )
+        if not presentation.connectionless:
+            await assert_valid_prover(
```
TimoGlastra, line 242: Althoug we can't check the connection, we _could_ check whether the proof request is valid in terms of the schema check. the assert_valid_prover does two checks

app/generic/verifier/verifier.py
morrieinmaas: ah alright nice didnt know that - will implement that then

app/generic/verifier/verifier.py
morrieinmaas: yes

TimoGlastra (approved):

PR #271, opened 2022-11-11
feat: auto remove v1 and v2 prefixes

### LINKED ISSUE #268: Remove v1-/v2- prefix for attachments in the connections/oob/create-invitation call
In the cloud api we return the proof id with a v1- or v2- prefix, however we do not strip it in the attachments array of the oob invitation call. We should strip the v1-/v2- prefix

```
"attachments": [{"id": "v2-65e023a5-88f1-4ed4-b9ff-63c36f4eb837", "type": "present-proof"}]
```

Should be sent to ACA-Py as:

```
"attachments": [{"id": "65e023a5-88f1-4ed4-b9ff-63c36f4eb837", "type": "present-proof"}]
```

morrieinmaas:

closes #268


TimoGlastra (approved):
Nice 👍

TimoGlastra (approved):

PR #272, opened 2022-11-13
feat: add thread id and parent thread id to models

### LINKED ISSUE #269: Add the thread_id and parent_thread_id properties to the credentials and proofs models
Otherwise it's impossible to match a webhook against an out of band invitation that was received.

morrieinmaas:

closes #269


shared_models/shared_models.py
```diff
@@ -239,6 +241,7 @@ class PresentationExchange(BaseModel):
     ] = None
     updated_at: Optional[str] = None
     verified: Optional[bool] = None
+    thread_id: Optional[str] = None
```
TimoGlastra, line 244: I think presentation model can also have a parent thread id?

TimoGlastra, line 244: I thiught thread_id is not optional for most models, can you check?

morrieinmaas: yes will check. However, I made that optional to cover the case where this is None for strange/unknown reasons we've seen before in the webhooks.

morrieinmaas: hmm let me double check - I didn't see parent thread id in the presentation exchange records but I'll look again

morrieinmaas: yeah that was me - well spotted. Parent thread id is there. will add that in

morrieinmaas: There's no parent_thread_id in V20PresExRecord nor in V10PresentationExchange. As far as I see. Can we get/derive this in a straight forward way from somewhere?

morrieinmaas: at least in the cloudcontroller api this is optional

TimoGlastra: Ah it seems it takes it from the presentation request message: https://github.com/hyperledger/aries-cloudagent-python/blob/0.7.5/aries_cloudagent/protocols/present_proof/v2_0/manager.py#L286

TimoGlastra: Yeah true, but having an always defined property is a lot nicer to work with than an optional attribute, so if we can prevent it being optional that would have my preference

TimoGlastra: > at least in the cloudcontroller api this is optional

Ah okay, maybe keep it optional then. I think the cloud controller says a lot of properties are optional that in fact, aren't.

morrieinmaas: yeah indeed. maybe that's something to get to the bottom of (in the cloudcontroller) as well.

morrieinmaas: ah alright thanks. That's exclusive to v2?

TimoGlastra: I _think_ it's because ACA-Py doesn't mark them as required, so we would have to fix this in ACA-Py itself 🤷

morrieinmaas: interesting btw - if they're optional in aca-py itself maybe there is a reason these are Optional? Or it's just being too permissive in the models...

TimoGlastra: Don't think so

morrieinmaas: yeah just saw it's in both (potentially) available

morrieinmaas: thx for the pointer

shared_models/shared_models.py
```diff
@@ -278,76 +279,80 @@ def presentation_record_to_model(
             connection_id=record.connection_id,
             created_at=record.created_at,
             error_msg=record.error_msg,
-            protocol_version=PresentProofProtocolVersion.v2.value,
+            parent_thread_id=record.pres_request.id if record.pres_request else None,
```
TimoGlastra, line 282: The parentThreadId should be taken from the threadId from the message. You can take the parentThreadId by finding the first message in the exchange (first try proposal then request) and then extracting the parent thread id from the thread decorator.

See https://github.com/hyperledger/aries-rfcs/blob/main/concepts/0008-message-id-and-threading/README.md for reference

morrieinmaas: hey thx for the details! I've inspecting this but these values were missing for webhooks (logging them raw/untransformed). I'll see if I can somehow get them 

thx again. very helpful

morrieinmaas: for example here an entire record as received via webhook

```python
{ 'auto_present': None,
                             'auto_verify': False,
                             'by_format': { 'pres': None,
                                            'pres_proposal': None,
                                            'pres_request': { 'indy': { 'name': 'string',
                                                                        'non_revoked': { 'from': 0,
                                                                                         'to': 20},
                                                                        'nonce': '12345',
                                                                        'requested_attributes': { '0_speed_uuid': { 'name': 'speed'}},
                                                                        'requested_predicates': { },
                                                                        'version': '1.0'}}},
                             'connection_id': 'a3b6e06c-1ced-4105-9d8a-94cdab9b8674',
                             'created_at': '2022-11-15T12:33:45.968448Z',
                             'error_msg': None,
                             'initiator': 'external',
                             'pres': None,
                             'pres_ex_id': '6992b055-bc62-4690-b877-76a90200ae8b',
                             'pres_proposal': None,
                             'pres_request': { 'comment': None,
                                               'formats': [ { 'attach_id': 'indy',
                                                              'format': 'hlindy/proof-req@v2.0'}],
                                               'id': '4b8cb07c-78c0-4430-a586-94c22a9e1454',
                                               'request_presentationsattach': [ { 'byte_count': None,
                                                                                  'data': { 'base64': 'eyJyZXF1ZXN0ZWRfYXR0cmlidXRlcyI6IHsiMF9zcGVlZF91dWlkIjogeyJuYW1lIjogInNwZWVkIn19LCAicmVxdWVzdGVkX3ByZWRpY2F0ZXMiOiB7fSwgIm5hbWUiOiAic3RyaW5nIiwgIm5vbl9yZXZva2VkIjogeyJmcm9tIjogMCwgInRvIjogMjB9LCAibm9uY2UiOiAiMTIzNDUiLCAidmVyc2lvbiI6ICIxLjAifQ==',
                                                                                            'json_': None,
                                                                                            'jws': None,
                                                                                            'links': None,
                                                                                            'sha256': None},
                                                                                  'description': None,
                                                                                  'filename': None,
                                                                                  'id': 'indy',
                                                                                  'lastmod_time': None,
                                                                                  'mime_type': 'application/json'}],
                                               'type': 'did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/present-proof/2.0/request-presentation',
                                               'will_confirm': True},
                             'role': 'prover',
                             'state': 'request-received',
                             'thread_id': '4b8cb07c-78c0-4430-a586-94c22a9e1454',
                             'trace': False,
                             'updated_at': '2022-11-15T12:33:45.968448Z',
                             'verified': None}
```

lohanspies: @morrieinmaas if `thread_id` the parent thread id or the specific message `thread_id`? Can we include `parent_thread_id` as well?

morrieinmaas: @lohanspies the idea is to include both. Having some trouble finding the parent thread id in the webhooks though (don't want to/cannot always do an extra call to receive records)

lohanspies: @morrieinmaas @TimoGlastra still a bit confused as to our webhooks implementation vs the below:
```There are several pieces in the community -- Redis and Kafka instances of persistent queues (BC Gov and SICPA), the Redis caching that you mention (Indicio), deployment scripts for testing (e.g. docker-compose) and OpenShift/K8s.

@swcurran
All of this is available and has had lots of work put in. I think we need to put it all in Hyperledger with consistent guidance on how to deploy in common environments.  Great to get an initiative around this:

Resources:

https://github.com/bcgov/aries-acapy-plugin-redis-events
https://github.com/sicpa-dlab/aries-acapy-plugin-kafka-events
https://github.com/Indicio-tech/aries-acapy-cache-redis
https://github.com/hyperledger/aries-cloudagent-python/blob/main/RedisPlugins.md```

Can you maybe provide some insight?

morrieinmaas: @lohanspies will have to look into this. However, not what I meant with my comment above. In the raw webhooks (not our implementation) received I'm not getting the parent thread id in the request attach or anywhere in the proof record received via webhook (as far as I can see).

morrieinmaas: so it's for handling webhooks in two ways:

1. Receive and redistribute webhooks in between aca-py and receiver
2. Transform/format webhooks to format used by cloudapi
3. authenticate who can receive what webhooks

morrieinmaas: alright as discussed with @TimoGlastra, @lohanspies FYI, parent thread id for proofs is not in the hooks - seems to be an aca-py bug. Will implement with only thread_id for now and open an issue in aca-py to fix this.

TimoGlastra (approved):

PR #274, opened 2022-11-14
feat: connectionless credential offer

### LINKED ISSUE #270: Add option to create a credential offer not bound to a connection
So like send-offer but then without connection_id like we already have with proof requests

morrieinmaas:
closes #270


app/tests/e2e/test_issuer.py
```diff
@@ -90,6 +89,73 @@ async def test_send_credential(
     assert_that(records).extracting("protocol_version").contains("v1", "v2")
 
 
+@pytest.mark.asyncio
+async def test_create_offer(
```
TimoGlastra, line 93: Can we add an e2e tests that creates an offer, creates a connectionless oob invitation, and the other agent accepting the oob inviation, and then waiting for the offer-received webhook event for credentials?

It's quite a complex flow that would be good to have a test for

morrieinmaas: I guess we can

morrieinmaas: isn't there a test for this in oob e2e tests already?

morrieinmaas: ah that one isn't connectionless nvm

TimoGlastra (approved):

PR #285, opened 2022-11-24
feat: little improvement for groups plugin

morrieinmaas:
Override the get wallets endpoint in the controller instead of using httpx client

This has the advantage of keeping types and auth

### DIFF
--- a/app/admin/tenants/tenants.py
+++ b/app/admin/tenants/tenants.py
@@ -16,6 +16,7 @@ from fastapi import APIRouter, Depends
 from pydantic import BaseModel
 from uplink import (
     Consumer,
+    Query,
     get,
     returns,
 )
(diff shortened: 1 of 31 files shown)


app/admin/tenants/tenants.py
```diff
@@ -207,28 +210,29 @@ async def get_tenants(
...
+                group_id=group_id,
+            )
+
+        @returns.json
+        @get("/multitenancy/wallets?group_id={group_id}")
+        def __get_wallets(self, *, group_id: str = None) -> WalletListWithGroups:
+            """Internal uplink method for get_wallets"""
+
+    aries_controller.multitenancy = MultitenancyApi(
+        base_url=aries_controller.base_url, client=aries_controller.client
```
TimoGlastra, line 228: Maybe we can do this when we create the aries controller instance?

morrieinmaas: see [this reply](https://github.com/didx-xyz/aries-cloudapi-python/pull/279#discussion_r1031640778)

morrieinmaas:
changes already merged in other PR

PR #292, opened 2022-11-30
docs: add example javascript websocket listener

morrieinmaas:

### DIFF
--- a/docs/Webhooks.md
+++ b/docs/Webhooks.md
@@ -68,15 +68,50 @@ docker logs --follow $(docker ps -f name="governance-webhooks-web" | awk 'FNR ==
 
 ### Non-Python options
 
-Listening to webhooks the subscriber way (not long polling via http) is pnot limited to the python example given above. In fact, all one needs is a websocket-based RPC client. 
+Listening to webhooks the subscriber way (not long polling via http) is pnot limited to the python example given above. In fact, all one needs is a websocket-based RPC client.
 
-Here are two examples:
+Here are two examples from the terminal:
 
 ```bash
 websocat -E --text ws://127.0.0.1:3010/pubsub exec:'{"request": {"method": "subscribe", "arguments": {"topics": ["proofs", "endorsements", "oob", "out_of_band", "connections", "basic-messages", "credentials"]}}}'
 ```
-or 
+
+or
+
 ```bash
 wscat -c ws://127.0.0.1:3010 -x '{"request": {"method": "subscribe", "arguments": {"topics": ["proofs", "endorsements", "oob", "out_of_band", "connections", "basic-messages", "credentials"]}}}' -w 99999
 ```
-How this works is that either procedure instantiates a client connecting to the websocket endpoint exposed via the webhooks container (*NOTE:* You might have to change the uri according to your setup of the webhooks relay). Both examples do pretty much the same. However, [Wscat](https://github.com/websockets/wscat) is written in Javascript whereas [websocat](https://github.com/vi/websocat) is implemented in Rust. Both examples are given to illustrate that it really does not matter what language one wishes to implement a listener in. After having established a connection to the exposed endpoint the `exec:` parameter and `-x` flag mean execute. Execute in this case refers to sending the JSON payload to the webhooks relay. It requests the endpoint to add the connection as a subscriber to the topics array of the arguments key. You can pass any arguments supported by the webhooks relay (see above). Passing an empty array under topics means 'end the subscription'. By Adding the `wallet_id` in the header is the way to only receive hooks for a specific wallet. 
+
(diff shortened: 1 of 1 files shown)


docs/Webhooks.md
```diff
@@ -68,15 +68,50 @@ docker logs --follow $(docker ps -f name="governance-webhooks-web" | awk 'FNR ==
...
+
+#### Javascript example
+
+Here's a simple example in javascript using a single dependency (`ws`):
+
+```javascript
+Object.assign(global, { WebSocket: require("ws") });
```
TimoGlastra, line 92: Why do we need this?

docs/Webhooks.md
```diff
@@ -68,15 +68,50 @@ docker logs --follow $(docker ps -f name="governance-webhooks-web" | awk 'FNR ==
...
+
+ws.on("message", (data) => {
+  console.log(data.toString());
+});
+```
+
+In order to use this:
+
+- Install `ws` dependency with
+  - `npm i ws` OR `yarn add ws` (you can add the --save-dev or -D flags to not install globally)
```
TimoGlastra, line 115: `--save-dev` / `-D` install it as a dev dependency. I you do `npm i -g` or `yarn global add` it will add it as a global dep

docs/Webhooks.md
morrieinmaas: @TimoGlastra, @lohanspies requested this.

morrieinmaas: @lohanspies it might be nice to make issues from this instead of messaging via slack/email. that way we can make it more transparent for Timo or anyone else to follow why certain things happen (even though messaging directly is quicker)

TimoGlastra: I meant the `Object.assign(global, { WebSocket: require("ws") });` 😄

lohanspies: @TimoGlastra @morrieinmaas BTW - we are having endless headaches trying to use Javascript with the `fastapi_websocket_pubsub` package being used in the code. I have spoken to the creator of the `fastapi_websocket_pubsub` library, and he said that you would need to build your own RPC and Pub/Sub client in Javascript to work with this package. The example @morrieinmaas gave works and then also doesn't work.

We have two options here... either we switch to a well-supported implementation that will enable any WebSocket clients to connect (JS, C#, Rust etc.), or we will need to develop custom clients for each framework able to listen to WebSocket events.

We are also not the only people with this problem - https://github.com/permitio/fastapi_websocket_pubsub/discussions/47#discussioncomment-3488961

i.e. We will need to make a call here as it is creating endless headaches for us trying to build solutions on top of Cloud API where we cannot robustly subscribe and receive WebSocket events.

Thoughts??

lohanspies: @morrieinmaas the example you gave us can subscribe to the WebSocket and receive events, however, the events are received intermittently and not in a very stable/usable way.

Here is the comment from the creator of the library we use:

My question:
```
I will post it in the discussion.
However it seems like when using this ws library cause some issues with your library.
As if the messages are delayed and or fail to deliver.
Need to figure out what is going on.
```

His response:

```
The answer is very simple, and I actually give it in [the thread here that I shared with you](https://github.com/permitio/fastapi_websocket_pubsub/discussions/47#discussioncomment-3488961) (Sorry for being a douche and quoting myself):
In this case for example you're ignoring the fact that the subscribe function on the pub/sub server expects the RPC client on the other side to implement the [notify method](https://github.com/permitio/fastapi_websocket_pubsub/blob/49b8dfc7eac055d264ea492d559cee975d5574a6/fastapi_websocket_pubsub/rpc_event_methods.py#L86). Without the notify method sending an RPC response back, which the [RPC server is waiting for](https://github.com/permitio/fastapi_websocket_rpc/blob/0edd769c58dc3df47f222b5133ac5606ef53c2f7/fastapi_websocket_rpc/rpc_channel.py#L263) the server would hang until a configurable timeout (which is forever by default).

 You can see here the [RPC code sending a response](https://github.com/permitio/fastapi_websocket_rpc/blob/0edd769c58dc3df47f222b5133ac5606ef53c2f7/fastapi_websocket_rpc/rpc_channel.py#L197), to implement the response you should send back a JSON object of this [format](https://github.com/permitio/fastapi_websocket_rpc/blob/0edd769c58dc3df47f222b5133ac5606ef53c2f7/fastapi_websocket_rpc/schemas.py#L16):
 ```


```
{
 "response":  {
      "result": "the return value",  
      "result_type": "the return value type name",
      "call_id": "the UUID that we sent in the request"
  } 
}
```

Comment (morrieinmaas) [reply docs/Webhooks.md:92]: Hey @lohanspies, so I think there is some confusion around this - mainly two-fold:

1. The direct subscription to the webhooksurl:3010 was/is meant to be mainly consumed by the cloudapi project internally. This part was not meant to be exposed to the www and consumed by the client as we agreed on in previous discussions. 
2. The cloudapi:8000 (if you look at the swagger for instance) exposes the webhooks via http over the cloudapi part thats meant to be exposed to www. That was partly also because at the time there was a) no client part implemented yet and b) folks implementing the client were somewhat reluctant about using (web)sockets. 

There is several ways to go about this.
1. Expose the webhooks handler/relay directly (change of intention).
2. Also provide websockets/subscribe via cloudapi www endpoint (port 8000 in current setup)

Regarding 1.
There needs to be some work done around security in this case but the main part "of it doesn't really work" as you mentioned above can be solved from the answer of the author you cite above - extend the minimal example I gave to send a response:

```javascript
ws.on("message", (data) => {
  console.log(data.toString());
  // send the response as stated by library author
  ws.send({
   "response":  {
        "result": "the return value",  
        "result_type": "the return value type name",
        "call_id": "the UUID that we sent in the request"
    } 
  });
});
```

Regarding 2. 
* You can poll the :8000/webhooks endpoint using http requests reliably and with auth (per wallet.
* If websockets is desired this indeed needs some more work but certainly possible.

Hope that helps is useful. Let me/us know which option you prefer or whether you see another option(s).

PR #303, opened 2023-04-21
fix/improve exception handling

ff137:
Stacktrace only prints when prod == False. prod is true by default


app/facades/__init__.py
wdbasson, line 1: What are these empty files?

wdbasson (commented):
@ff137 how do you set `prod == False`? Is it possible to do this with an env var?

app/facades/__init__.py
ff137: Empty `__init__.py` files are necessary for the interpreter to recognize imports of modules in the directory. It's not strictly necessary in newer version of Python, but it's considered best practice & it's primarily for imports to resolve correctly in vscode

ff137: with regards to setting `prod == False` , there seems to be an environment variable called `prod`. If it doesn't exist, defaults to true, as follows: `prod = strtobool(os.environ.get("prod", "True"))`.

If I search for "prod" it doesn't seem to be set in any of the docker files... so we should make a todo for that - either create new env var or point to a different one.

wdbasson (approved):
LGTM

ff137:
Replicated from clean branch in https://github.com/didx-xyz/aries-cloudapi-python/pull/308

PR #306, opened 2023-04-25
Didx 272 merge workflows into upstream

wdbasson:
Implements #304

continuous build pipeline:
- Build, deploy and test against private EKS cluster
- Self-hosted runner
- Clean startup

docker-compose enhancements and improvements:
- ability to reference pre-built images by passing `--no-build` set `REGISTRY` and `IMAGE_TAG`
- handling inter-service startup dependencies
- removed sleep hacks

dockerfile enhancements:
- Dockerfile `ENTRY_POINT` instead of `CMD` to allow overriding args with k8s see: https://jamesdefabia.github.io/docs/user-guide/containers/

test improvements/workaround:
- `./app/facades/acapy_ledger.py`: accept_taa_response hack
- source `./app/tests/util/constants.py` as env vars allowing override
- `./app/generic/definitions.py` fix to check for 405 response code
- `./app/tests/e2e/test_connections.py` improvement to assert both http/https


.github/workflows/continuous-deploy.yml
```diff
@@ -0,0 +1,390 @@
...
+              diff upgrade "${project}" \
+              --install \
+              --set image.tag=${{ needs.build.outputs.image_version }} \
+              -f ./helm/Chart/values/dev/${project}.yaml \
+              ./helm/Chart | tee -a diff_output.txt
+          done
+
+      - name: Quiet diff # Display lines with actual changes and lines indicating the service or component that has changed
+        run: |
+          grep -E '^[-+]\s|has changed:$' diff_output.txt | awk '{$1=$1};1'
```
rblaine95, line 245: That's very clever, I like that

app/generic/definitions.py
lohanspies, line 1: Not sure about this change

scripts/env.sh
lohanspies, line 1: Think these values should not be public

wdbasson: Sensitive information now exposed as GitHub secrets populated with Terraform.

rblaine95 (approved):

app/tests/e2e/test_connections.py
```diff
@@ -21,7 +21,7 @@ async def test_create_invitation(
     assert_that(invitation["invitation"]).is_instance_of(dict).contains(
         "@id", "@type", "recipientKeys", "serviceEndpoint"
     )
-    assert_that(invitation["invitation_url"]).contains("http://")
+    assert_that(invitation["invitation_url"]).matches(r"^https?://")
```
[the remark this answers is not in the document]
ff137: From match method: `Match is partial unless anchored`, so not a problem

app/generic/definitions.py
```diff
@@ -474,7 +474,7 @@ async def create_schema(
         # If status_code is 405 it means the schema already exists in the trust registry
         # That's okay, because we've achieved our intended result:
         #   make sure the schema is registered in the trust registry
-        if error.status_code != 400:
+        if error.status_code != 405:
```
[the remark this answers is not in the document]
wdbasson: Reverted to 400. (But still probably worth adding some more exception handling/logging.)

PR #308, opened 2023-05-03
Fix/remove stacktrace from jsonresponse

ff137:
Created new PR to remove changes included from branch `feat/bump-acapy-080`

Original PR: https://github.com/didx-xyz/aries-cloudapi-python/pull/303

Stacktrace only prints when prod == False. prod is true by default


wdbasson (approved):
LGTM

Just add a todo to figure out how to enable the stack trace with an env var so that we can override it when needed without having to hardcode it.

ff137:
Todo created 👍

Issue #259, opened 2022-10-17
webhooks auth


morrieinmaas:
Currently the webhooks handler has no auth mechanism. This is a design choice as it is intended to be consumed only by the cloudapi setup internally. 

However, one might want to listen to the filter webhooks via websocket subscriber from a client app (external to cloudapi). If so, some form of auth should be implemented.

Is this a desired feature/change?

Suggestion: 

Auth should cover at least the things:

1. Auth per wallet. A subwallet (client) should only be able listen to its own webhooks corresponding to its wallet id. It should NOT be able to listen to other wallets' hooks nor all webhooks
2. A webhook subscriber for a given wallet should have to pass its Bearer token for authentication.[^1]
3. The exposed websocket endpoint should be secured (wss instead of ws - OUT OF SCOPE  - infra)

[^1]: In order to keep the webhook relay stateless and not having to maintain an auth database it could do a call againt the aca-py instance with the provided header Bearer token to query something for the subwallet (provided wallet_id). If the call returns with status 200 (and not with 401 unathorized) the token is valid. Since the listener connection is keep-alive and not http long poll this should not flood the network. Still feels slightly hacky so other ideas welcome.

TimoGlastra:
I think the approach you describe in the footnote is completely fine @morrieinmaas, and also what we're doing for the webhook endpoints in the cloud API at the moment I believe 👍

ff137:
Auth has since been implemented. The tenant/ and tenant admin/ sse, websocket and webhook endpoints all authenticate that calls are made by admins or relevant tenant_jwt's

PR #309, opened 2023-05-04
Improve webhooks + listener implementation

ff137:
The webhooks + listener implementation (`app/webhook_listener.py`) is laden with code smells:
- poor function names (`on`/`off`; `on_webhook` in listeners and `_on_webhook` in Webhooks class; `_listeners` is the name for the list of callback functions) with no docstring descriptions
- `sys.exit()` being called on timeout errors
- listener is implemented as an `async def start_listener()` which returns two async functions when called: `return wait_for_event_with_timeout, stop_listener`. This should be moved to its own class, providing instance specific access to these methods.
- `while True` to read from an async queue. By default, `get()` from an async queue will wait indefinitely if it's empty. There is also a duplicate method to introduce timeout to this loop
- the match payload to filter_map method:
  ```python
              match = all(
                  payload.get(filter_key, None) == filter_value
                  for filter_key, filter_value in filter_map.items()
              )
  ```
  will always return true when the key in filter_map doesn't exist in payload and if filter_map value is searching for None.
- and some more minor code smells.

This PR refactors `webhook_listener.py` into `webhooks.py` and `listener.py` for better maintainability, fixing code smells, and updating usage of these instances to work as prior.

Further notes:
- Webhooks is currently implemented as a static class with static methods and class-level variables. Therefore, it does not manage any instance specific state, and will have a shared set of listeners with a single Webhooks instance for the entire application. Currently the Webhooks instance manages a single PubSubClient with topics set to `'ALL_WEBHOOKS'`. I'll need some feedback about how to improve this, but creating instance specific webhooks may yield better maintainability + scalability.
- It may be worth adding an auth mechanism to listeners, as noted here: https://github.com/didx-xyz/aries-cloudapi-python/issues/259
- Test mocking in `test_onboarding.py` for `test_onboard_issuer_public_did_exists` seems to have been done incorrectly. Mock event listeners were created, but there's a mismatch in parameter values. Adding to fix list

### DIFF
--- a/app/admin/tenants/onboarding.py
+++ b/app/admin/tenants/onboarding.py
@@ -174,7 +174,8 @@ async def onboard_issuer(
                 }
             )
         except TimeoutError:
-            raise CloudApiException("Error creating connection with endorser", 500)
+            raise CloudApiException(
+                "Error creating connection with endorser", 500)
         finally:
             listener_transaction.stop()
 
@@ -222,7 +223,8 @@ async def onboard_issuer(
                 }
             )
         except TimeoutError:
-            raise CloudApiException("Error creating connection with endorser", 500)
+            raise CloudApiException(
+                "Error creating connection with endorser", 500)
         finally:
             listener_transaction.stop()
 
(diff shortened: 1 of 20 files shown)


lohanspies:
@ff137 LGTM

PR #311, opened 2023-05-11
Add SSE to webhooks

ff137:
Add SSE capability to webhooks server.

SSE stream will stay open indefinitely and so current implementation will yield the cached events and suspends after 1 second.

Also:
- organised webhooks routes and dependencies in subpackages 
- refactored following routes:
  - `/{wallet_id}` -> `/webhooks/{wallet_id}` 
  - `/{topic}/{wallet_id}` -> `/webhooks/{wallet_id}/{topic}`
- organised and refactored `/shared_models` -> `/shared` and moved some more common methods
- wrote test for webhooks + SSE. Added to continuous build pytests (previously not set up to test /webhooks as well)
- fixed `connections` fixtures to use data classes
- fixed all flake8 warnings
- using custom APIRouter to automatically duplicate route definitions with & without trailing slash, avoids redirects

Remaining to-do:
- add auth to routes
- improve SSE to suspend stream when desired event is returned
- refactor webhooks listeners in /app to use SSE instead

N.B. currently includes `echo "0.0.0.0 wh-redis" | sudo tee -a /etc/hosts` hack in continuous-build for redis to resolve.


ff137:
@wdbasson The tests in `/webhooks` will fail due to `wh-redis` not being resolved (`aioredis.exceptions.ConnectionError: Error -3 connecting to wh-redis:6379. -3`). Quick local fix for me was adding `127.0.0.1 wh-redis` to my /etc/hosts. 

Thought it was my docker-network config on WSL, but I also needed the fix here, by adding to continuous-build: `echo "0.0.0.0 wh-redis" | sudo tee -a /etc/hosts`... but of course it's preferable not to hack the /etc/hosts file in this way. Any idea how to resolve this more elegantly?

PR #312, opened 2023-05-11
Art/update docs

ff137:


ff137 (commented):
@lohanspies docs now match with the changes in ssi-dev-portal

PR #314, opened 2023-05-12
Upgrade pytests: use anyio instead of pytest-asyncio

ff137:
`pytest-asyncio` is no longer maintained. `anyio` is the better replacement, and is used by other dependencies such as `fastapi`.

Refactoring primarily consists of replacing the test decorates `@pytest.mark.asyncio` with `@pytest.mark.anyio`, and then defining an anyio backend in the conftests.py

Notes:
- pushing all pytest dependencies (`pytest`, `pytest-cov`, `mockito`, `anyio`) to latest versions.
- `/webhooks/tests.py` were never set up to run. Only contains 1 test, which failed as it was written. Improved and fixed 
- some failing tests seems to be contingent on other PRs being merged


ff137:
Only struggle at the moment is getting all async resources to close nicely during teardown. Currently we only have errors during teardown: `RuntimeError: Event loop is closed`, meaning that fixtures are calling async tasks during shutdown.

ff137:
Progress. 

I've managed to fix the teardown errors. They seemed to occur as a result of numerous await statements being queued after the yield statement in the different pytest fixtures. For example, tenants are deleted by the admin client during teardown, and other clients are closed.

These async calls seem to fail during teardown when there are too many of them ... I'm sure there is configuration to work around this, but instead I've opted to change the fixture scope from `module` to `function`, meaning that pytest fixtures are created for, and destroyed after, every function test, instead of after every module's tests. This makes sense to me, because the tests are already mostly set up to be self-contained; only a few of them rely on other tests being run before them.

This makes the tests much slower, because relevant clients are spun up and torn down for each test function.

On the bright side, there are no more errors. Only 2 tests fail now, because they are not self-contained. So, further work is to refactor these tests to indeed be self-contained, and perhaps to re-introduce module level clients, where it makes sense.

ff137:
The two failing tests are as follows:
```
FAILED app/tests/e2e/test_issuer.py::test_get_records - assert []
FAILED app/tests/e2e/test_verifier.py::test_get_credentials_for_request - IndexError: list index out of range
```

`test_get_records` only does one thing: `get "/credentials"`, asserting that some records exist, created from previous tests. This test can be removed, because the `get "/credentials"` functionality is already being asserted in `test_send_credential_oob_v1`, `test_send_credential`, and `test_send_credential_request`.

`test_get_credentials_for_request`: seems to me that a fixture (`issue_credential_to_alice`) was missing for this test, or at least it's required to make it self-contained.

ff137:
@wdbasson when this branch is merged, then parallelising tests will be possible. 

They can now be run individually at a function level, or perhaps we can just run the slowest .py files on their own runner.

PR #318, opened 2023-05-15
Upgrade fastapi & starlette dependencies

ff137:
Notes:
- The only tests that broke with the dependency upgrades were in `/trustregistry/tests`
- These tests were using a test db session to validate crud methods for actors + schema tables, and broke with starlette 0.15 - could not see why.
- I upgraded all these tests to rather be performed against the actual trust registry db. All tests delete any actors/schemas they create, so should not make changes to trustregistry.
- deleted `test_crud` as it was still dependent on a test db, and all crud methods are already validated by `test_actor` and `test_schema`


ff137:
Managed to find the breaking upgrade: fastapi 0.69.0 uses starlette 0.15. Using those versions or newer results in database errors

PR #319, opened 2023-05-16
Remove deploy job from `continuous-build.yml` and update CodeQL workflow

ff137:
The purpose of the current deploy job is unclear, and is currently failing. Resolution is to remove it for the meantime.

Made the duplicate PR to test out the failing CodeQL analysis


morrieinmaas:
@ff137 the purpose of the deploy job was to deploy it on a server that @lohanspies hosted somewhere a while ago so he can play around with the latest version. It was basically the dev setup hosted on www.

ff137:
@morrieinmaas Cool, thank you for the feedback! That makes sense, means it can be safely removed.

PR #325, opened 2023-05-17
Chore: improve error logging and exception messages

ff137:


app/generic/verifier/facades/acapy_verifier_v1.py
```diff
@@ -44,44 +36,52 @@ async def create_proof_request(
 
     @classmethod
     async def get_credentials_for_request(cls, controller: AcaPyClient, proof_id: str):
+        pres_ex_id = pres_id_no_version(proof_id=proof_id)
```
ff137, line 39: This is moved outside of the try, because it can raise a ValueError which should not be caught as general Exception

app/generic/verifier/facades/acapy_verifier_v2.py
```diff
@@ -1,73 +1,72 @@
...
+                                         CreateProofRequest,
+                                         RejectProofRequest, SendProofRequest)
+from shared_models import PresentationExchange, pres_id_no_version
+from shared_models import presentation_record_to_model as record_to_model
 
 logger = logging.getLogger(__name__)
 
 
 class VerifierV2(Verifier):
+    @classmethod
```
ff137, line 22: Rerranged this method just to line up with order in VerifierV1 implementation

PR #324, opened 2023-05-17
Run tests in parallel with `pytest-xdist`

ff137:

PR #337, opened 2023-05-29
Implement `RichAsyncClient`, and more bulk refactoring to tests

ff137:
As precursor to parallelising the tests (#324), some refactoring is needed to improve the structuring of pytest fixtures.

For one thing, this pattern needs to happen a lot:
```
response = client.post(url, json=*)
if response.is_error:
  raise Exception("Something went wrong")
```
This becomes incredibly laborious to add for every single post/get, but it's necessary for proper error handling. In the case of the pytests, it's necessary for debugging.

Solution: create a RichAsyncClient that extends the AsyncClient, and add custom error handling for its post/get methods.

This just requires some refactoring for tests that assert errors being raised.

While I'm at it, I'm also trying to move away from using json everywhere, in the request body and in the responses that get passed around, because then you need to do additional validation that the json matches your expected object, and you need to subscript "unknown" variables out of the json. We have the models for a reason, so lets use them! So I made a `parse_with_error_handling` method to parse our responses into the expected object, and then we can say `tenant.access_token`, knowing that the field must exist, instead of checking `if not tenant["access_token"]: raise Exception`.

So, big refactoring to get all tests to pass again, and then parallelising will be easier.


ff137:
Pytest fixtures for AcapyClients, AsyncClients, Wallets, Connections, Invitations, etc, were previously disorganized across `client_fixtures`, `member_personas`, `ecosystem_personas`. For maintainability, there should be some common thread for where different fixtures types are defined. To this end, I moved all acapy_client fixtures to the same file, same for all async_client fixtures, all wallet fixtures, and all connections fixtures. Maintainability is now through the roof.

It's now much easier to see what sort of consistency exists between fixtures, easier to identify bugs, and easier to optimize. For one thing, I could immediately identify methods that were defined as fixtures that didn't need to be. I can also see some inconsistency in the patterns used in fixtures of the same type. And we also resolve a load of Codacy issues w renaming methods out of scope.

app/generic/messaging.py
```diff
@@ -22,7 +22,7 @@ class TrustPingMsg(BaseModel):
     comment: str
 
 
-@router.post("/send-message", status_code=204)
```
ff137, line 25: @lohanspies I found that `send-message` route had a custom response status code `204`, causing it to deviate from the acapy spec. For consistency I'm reverting it back to `200`.

PR #377, opened 2023-07-10
Feat/create actor checks

cl0ete:
Add exception handling for actor creation if actor already exits 
Update tests


trustregistry/crud.py
```diff
@@ -51,26 +52,53 @@ def get_actor_by_id(db_session: Session, actor_id: str) -> db.Actor:
...
+                f"Bad request: The requested actor name: `{actor.name}` already exists in database."
+            )
+
+        elif f"actors.did" in constraint_violation:
+            if f"actors.didcomm_invitation" in constraint_violation:
+                bound_logger.info(
+                    "Bad request: The requested actor DIDComm invitation already exists in database."
+                )
+                raise ActorAlreadyExistsException(
+                    f"Bad request: The requested actor DIDComm: `{actor.didcomm_invitation}` invitation already exists in database."
```
ff137, line 93: Here there's a missing "invitation" string in the message.

Also - this didcomm_invitation can be a HUGE string. I think it's fair to exclude it here, and just say: "An actor with this didcomm invitation already exists."

I think each of the "The requested actor ... already exists" strings can rather just say "An actor with ...  already exists"

trustregistry/crud.py
```diff
@@ -208,6 +236,9 @@ def delete_schema(db_session: Session, schema_id: str) -> db.Schema:
 class ActorAlreadyExistsException(Exception):
     """Raised when attempting to create an actor that already exists in the database."""
 
+    def __init__(self, message):
+        self.message = message
```
ff137, line 240: Optional: With this custom def __init__, you can also override the string method:
```
    def __str__(self):
        return self.message
```

This is allows you to use `e` in a f-string context, without needing to say strictly call e.message

trustregistry/tests/test_actor.py
```diff
@@ -12,6 +12,34 @@
...
+    "didcomm_invitation": "string",
+    "did": "did:key:string4",
+}
+dup_id_actor = {
+    "id": "darth-vader",
+    "name": "Darth Vader5",
+    "roles": ["issuer", "verifier"],
+    "didcomm_invitation": "string5",
+    "did": "did:key:string5",
+}
```
ff137, line 42: It feels like there should be a way to do this without the duplicate definitions ...
Like, you can define a def generate_actor(), where it returns a dict with keys: id, name, didcomm_invitation, did; where they are all randomly generated using our helper method (from app.tests.util.string import random_string)

So you can say dup_did_actor = generate_actor()
dup_did_actor["did"] = predefined_string

This way it's less duplication and more clear which key is being recycled

trustregistry/crud.py
```diff
@@ -51,26 +52,53 @@ def get_actor_by_id(db_session: Session, actor_id: str) -> db.Actor:
...
+                raise ActorAlreadyExistsException(
+                    "Bad request: An actor with DIDComm invitation already exists in database."
+                )
+            else:
+                bound_logger.info(
+                    "Bad request: An actor with DID already exists in database."
+                )
+                raise ActorAlreadyExistsException(
+                    f"Bad request: An actor with DID: `{actor.did}` already exists in database."
+                )
```
ff137, line 101: `elif` has no final `else` case - so could swallow `IntegrityError` that doesn't match on checks. Should be one final else to raise a generic "Unique constraint failed" message, in case specific if checks don't match.

I think the nested one:
```py
       elif "actors.did" in constraint_violation:
            if "actors.didcomm_invitation" in constraint_violation:
```

can just be replaced with `elif "actors.didcomm_invitation" in ... :` followed by `elif "actors.did" in ...` and then else: Generic log/response (with an error log to indicate that a constraint failed which isn't manually handled)

ff137 (approved):
Looks good :+1:

PR #379, opened 2023-07-14
Delete wallet if actor exists

cl0ete:
- Added get actor by name from trustregistry
- Added assert_actor_name: check if actor exists with given name
- Check if actor exists before creating tenant
- Raise 409 if actor exists
- Delete wallet if `onboard_tenant` or `register_actor` fails 
- Update test


app/admin/tenants/tenants.py
```diff
@@ -75,41 +76,69 @@ async def create_tenant(
...
-                didcomm_invitation=onboard_result.didcomm_invitation,
-            )
+    actor_exists = await assert_actor_name(body.name)
+
+    if actor_exists:
+        bound_logger.info("Actor exists can't create wallet")
+        raise HTTPException(
+            409, f"Can't create Tenant. Actor with name: {name} already exists."
         )
+    else:
```
ff137, line 89: Stylistic: `else` can be dropped, considering `if` case would have returned / raised an exception

app/admin/tenants/tenants.py
```diff
@@ -75,41 +76,69 @@ async def create_tenant(
...
+            bound_logger.info("HTTP exception: {}", http_error.detail)
+            if wallet_response:
+                bound_logger.info("delete wallet")
+                await admin_controller.multitenancy.delete_wallet(
+                    wallet_response.wallet_id
+                )
+            raise http_error from http_error
+
+        except Exception as e:
+            bound_logger.info("Something went wrong")
```
ff137, line 135: Verbose logs are preferable: "Could not register actor; deleting wallet" and "An unhandled exception occurred" - with either printing `e`, or using `bound_logger.exception` (automatically prints stack trace).

app/admin/tenants/tenants.py
```diff
@@ -75,41 +76,69 @@ async def create_tenant(
...
-                id=wallet_response.wallet_id,
-                name=name,
-                roles=roles,
-                did=onboard_result.did,
-                didcomm_invitation=onboard_result.didcomm_invitation,
-            )
+    actor_exists = await assert_actor_name(body.name)
+
+    if actor_exists:
+        bound_logger.info("Actor exists can't create wallet")
```
ff137, line 85: Nitpick: actor _name_ already exists (for the logs), and variable can be `actor_name_exists`

app/admin/tenants/tenants.py
```diff
@@ -75,41 +76,69 @@ async def create_tenant(
...
+        raise HTTPException(
+            409, f"Can't create Tenant. Actor with name: {name} already exists."
         )
+    else:
+        try:
+            async with get_tenant_admin_controller() as admin_controller:
+                wallet_response = await admin_controller.multitenancy.create_wallet(
```
ff137, line 92: Considering we first checked if actor name exists, there should not be a log before create_wallet to indicate something like actor name is unique and now wallet is being created.

app/admin/tenants/tenants.py
```diff
@@ -75,41 +76,69 @@ async def create_tenant(
...
+                    )
+                )
+        except HTTPException as http_error:
+            bound_logger.info("HTTP exception: {}", http_error.detail)
+            if wallet_response:
+                bound_logger.info("delete wallet")
+                await admin_controller.multitenancy.delete_wallet(
+                    wallet_response.wallet_id
+                )
+            raise http_error from http_error
```
ff137, line 132: This would be redundant, as `from` is used to indicate "new exception being raised is from this original exception".
For brevity, can actually just be `raise`, because that automatically raises most recently caught exception. Similar for the `raise e from e` in next catch - `e` would not need to be assigned in that case, but `http_error` is used in log of the first catch, so that can remain assigned.

app/facades/trust_registry.py
```diff
@@ -144,6 +144,47 @@ async def actor_by_did(did: str) -> Optional[Actor]:
...
+    elif actor_response.is_error:
+        bound_logger.error(
+            "Error fetching actor by name. Got status code {} with message `{}`.",
+            actor_response.status_code,
+            actor_response.text,
+        )
+        raise TrustRegistryException(
+            f"Error fetching actor by name: `{actor_response.text}`",
+            actor_response.status_code,
+        )
```
ff137, line 182: This exception will bubble up to the client, because the `assert_actor_name` call is not in a try-catch in the create-tenant method. In case some other network error occurs (500), it'll be better to wrap `actor_exists = await assert_actor_name(body.name)` in a try-catch, and handle `except TrustRegistryException` by replacing the error message to the client with something more relevant to their original request of creating a tenant (because they won't understand "error fetching actor by name"). Instead we can raise CloudApiException, status 500, with "An error occurred when trying to register actor. Please try again". Not a great error message, but will be more informative on the off chance it occurs.

app/tests/facades/test_trust_registry.py
```diff
@@ -309,3 +309,32 @@ async def test_update_actor(mock_async_client):
...
+        didcomm_invitation="actor-didcomm-invitation",
+    )
+    mock_async_client.get = AsyncMock(
+        return_value=Response(status_code=200, json=actor)
+    )
+
+    assert await trf.assert_actor_name(name) is True
+
+    # test actor does not exists
+    not_actor = "Oduim"
```
ff137, line 332: Variable assignment only used once. Can instead call `trf.assert_actor_name("invalid actor name")` or something

trustregistry/registry/registry_actors.py
```diff
@@ -90,6 +90,19 @@ async def get_actor_by_id(actor_id: str, db_session: Session = Depends(get_db)):
     return actor
 
 
+@router.get("/name/{actor_name}")
+async def get_actor_by_name(actor_name: str, db_session: Session = Depends(get_db)):
+    bound_logger = logger.bind(body={"actor_name": actor_name})
+    bound_logger.info("Get request received: Get actor by name")
```
ff137, line 96: For consistency, route verbs are in all caps. i.e. GET request received

trustregistry/registry/registry_actors.py
```diff
@@ -90,6 +90,19 @@ async def get_actor_by_id(actor_id: str, db_session: Session = Depends(get_db)):
...
+    try:
+        actor = crud.get_actor_by_name(db_session, actor_name=actor_name)
+    except crud.ActorDoesNotExistException:
+        bound_logger.info("Bad request: Actor with name {} not found", actor_name)
+        raise HTTPException(status_code=404, detail="Actor not found.")
```
ff137, line 101: Exception message can be more verbose (like other routes): `f"Actor with name {actor_name} not found."`

trustregistry/crud.py
```diff
@@ -50,6 +50,22 @@ def get_actor_by_id(db_session: Session, actor_id: str) -> db.Actor:
...
+
+    if result:
+        bound_logger.info("Successfully retrieved actor from database")
+    else:
+        bound_logger.info("Actor name not found")
+        raise ActorDoesNotExistException
+
+    return result
+
+
```
ff137, line 68: Btw, not related to this function, but I can't add comment on the actual line:
```
        else:
            bound_logger.info("Bad request: {}", constraint_violation)
```
at the end of `create_actor`. 
I notice that does not raise an exception, in which case it would return None - in the unlikely case that constraint violation isn't matched by the other ifs. It should just return a generic "Bad request: Unique constraint violated - {}" with the constraint violation printed.

Also that final log message `bound_logger.info("Bad request: {}", constraint_violation)` should be error level, because it should never happen and means somethings changed.

ff137 (approved):

PR #402, opened 2023-07-26
Redis event revocation

cl0ete:
Add transformer for revocation webhook event
Add transformer for issuer_cres_rev webhook event


ff137 (approved):
Looks good :+1:

PR #415, opened 2023-08-04
Fix todos and address some security issues

cl0ete:
- Made long strings multi line
- escaped some brackets in SECURITY.md
- Logging a warning in `accept_proof_request` if `assert_valid_prover`  is not called. Looks like if proof request does not have a `connection_id` it is treated as a `oob` connection
- Removed a duplicate class `WalletRecordWithGroups` in `app/models/tenants.py` 
- Renamed locally scoped variable `cred_ex_id` to `cred_ex_id_b` not to confuse it with global `cred_ex_id` in `test_revocation_registry.py`

- Move `OnboardResults` class to `models/tenants.py`
- renamed instances of the Verifier V1/V2 interface from `prover` to `verifier`

### DIFF
--- a/app/routes/verifier.py
+++ b/app/routes/verifier.py
@@ -284,6 +284,8 @@ async def accept_proof_request(
                     prover=prover,
                     presentation=presentation,
                 )
+            else:
+                bound_logger.warning("No connection associated with proof, OOB?")
 
             bound_logger.debug("Accepting proof record")
             result = await prover.accept_proof_request(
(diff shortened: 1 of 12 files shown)


app/routes/verifier.py
```diff
@@ -284,6 +284,8 @@ async def accept_proof_request(
                     prover=prover,
                     presentation=presentation,
                 )
+            else:
```
ff137, line 288: In this situation, we're only asserting that it's a valid prover if there is a connection id. I feel like a to-do should still remain here, so we have a way of asserting a valid prover even if there's no connection associated with the proof.

ff137 (approved):

app/routes/verifier.py
ff137: I updated the log message. Not needed to resolve further

PR #423, opened 2023-08-15
Containerize tests

cl0ete:
Containerized the tests and output test logs to `test-outputs`.

Can now run tests with `./manage tests` or loop test with `./loop_tests`

`./loop_tests` will `./manage tests` wait for tests to exit then `./manage down` and then restart. 
`loop_tests` will loop 10 times if no arg is given or you can pass an arg like this `loop_tests 15` 

there is now a `summary.py` script in the scripts folder that will generate a summary of the test out-puts in `logs/containerized-tests` the summary will also be saved to `logs/containerized-tests`

the `manage test` script will also save the compose up logs to `logs/compose-up`


.gitignore
```diff
@@ -58,6 +58,7 @@ junit.xml
 .hypothesis/
 .pytest_cache/
 cover/
+logs/containerized-tests
```
ff137, line 61: Ah, I thought logs/ was already ignored, but I see it's covered by `*.log`
In this case, it seems we're saving an .xml file, which is already ignored. 
So the additional ignore shouldn't be needed, but we can instead just ignore the entire logs/ folder for clarity

loop_tests.sh
```diff
@@ -1,7 +1,23 @@
 #!/bin/bash
```
ff137, line 1: This .sh file can be moved to scripts/

ff137 (changes requested):
One last request - Ideally we'd want to keep the startup logs for a specific test run.

The pytest output is saving as: `pytest --junitxml="/tests/outputs/test_output_$(date +'%Y-%m-%d_%H-%M-%S').xml"`
In some cases where startup fails, all the tests may also fail. Then it would be helpful to also have a startup log with the same timestamp, for that run... 

The app runtime logs will already be saved, to investigate specific tests failing, but if startup fails, it would be great to have just the startup logs, to see what happens for all tests to fail -- maybe there's a way to truncate logs after a specific line is printed, like Application ready? Not sure how easy it is, but must be a way!

ff137 (approved):

PR #429, opened 2023-08-22
Add tests for issuing JSON-LD credentials with issuer did:sov and did:key bbs/ed25519

cl0ete:
- Refactor issuer tests.
- Add optional arg to create did endpoint `app/routes/wallet create_did()` and `app/services/acapy_wallet create_did()`
- add tests for jsonld for ed25519 and Bbs prooftypes 
- Added fixture to add actor with did:key to trust registry


app/tests/issuer/test_issuer.py
```diff
@@ -234,6 +236,7 @@ async def test_request_credential_x_no_schema_cred_def(
 
     v1_record.credential_definition_id = None
     v1_record.schema_id = None
+    v1_record.type = "indy"
```
ff137, line 239: If we set "indy" as the default type in the CredentialExchange definition, then setting it here won't be necessary

shared/models/topics/base.py
```diff
@@ -80,6 +81,7 @@ class CredentialExchange(BaseModel):
     ] = None
     # Thread id can be None in connectionless exchanges
     thread_id: Optional[str] = None
+    type: Optional[str] = None
```
ff137, line 84: Maybe we should make this required (`type: str`) and default its value to indy. Because I presume from the changes we're making that the type must be defined, and defaulting to indy makes it backward compatible

app/tests/issuer/test_issuer.py
cl0ete: Looks like `mock(CredentialExchange) ` does not inherit the default values of the model. Mr GPT had this to say: `No, when you use the mockito package to create a mock object, it does not automatically inherit default values from the actual Pydantic model. The purpose of creating a mock object is to simulate the behavior of the original object without involving its actual implementation, including default values.`

app/tests/e2e/issuer/test_issuer_ld_bbs.py
```diff
@@ -0,0 +1,382 @@
...
+
+    did = register_key_issuer
+
+    # Creating JSON-LD credential did:key
+    credential["connection_id"] = faber_connection_id
+    credential["ld_credential_detail"]["credential"]["issuer"] = f"{did}"
+    credential["ld_credential_detail"]["options"] = {
+        "proofType": "Ed25519Signature2018"
+    }
+    # Send credential must fail did:sov cant issue proofType: Ed25519Signature2018
```
ff137, line 175: did key + BBS can't issue Ed25519

app/tests/e2e/issuer/test_issuer_ld_bbs.py
```diff
@@ -0,0 +1,382 @@
...
+    alice_member_client: RichAsyncClient,
+    register_key_issuer: DidKey,
+):
+    alice_connection_id = faber_and_alice_connection.alice_connection_id
+    faber_connection_id = faber_and_alice_connection.faber_connection_id
+
+    did = register_key_issuer
+
+    # Creating JSON-LD credential did:key
+    credential["connection_id"] = faber_connection_id
```
ff137, line 90: credential = deepcopy(credential_)

app/tests/e2e/issuer/test_issuer_ld_bbs.py
```diff
@@ -0,0 +1,382 @@
...
+            issuanceDate="2021-04-12",
+            issuer="",
+        ),
+        options=LDProofVCDetailOptions(proofType=""),
+    ),
+).dict(by_alias=True, exclude_unset=True)
+
+
+@pytest.mark.anyio
+@pytest.mark.parametrize("register_key_issuer", ["bls12381g2"], indirect=True)
```
ff137, line 77: I think for readability it'll be better to have 2 separate fixtures: `register_issuer_did_key_bbs` and `register_issuer_did_key_ed25519`. Then we can have two dataclasses to represent the fixture types: DidKeyBbs and DidKeyEd25519.

At the moment, we have to scroll up to this parametrize method to see which signature type is being used. And it'll be clearer in the code to see `did = register_issuer_did_key_bbs`

app/tests/e2e/issuer/test_issuer_ld_bbs.py
```diff
@@ -0,0 +1,382 @@
...
+                "degree": {
+                    "type": "BachelorDegree",
+                    "name": "Bachelor of Science and Arts",
+                },
+                "college": "Faber College",
+            },
+            issuanceDate="2021-04-12",
+            issuer="",
+        ),
+        options=LDProofVCDetailOptions(proofType=""),
```
ff137, line 71: proofType can be set here, considering each test here is for bbs

app/tests/e2e/issuer/test_issuer_ld_bbs.py
```diff
@@ -0,0 +1,382 @@
...
+async def test_send_jsonld_request(
+    alice_member_client: RichAsyncClient,
+    faber_client: RichAsyncClient,
+    faber_and_alice_connection: FaberAliceConnect,
+    register_key_issuer: DidKey,
+):
+    alice_connection_id = faber_and_alice_connection.alice_connection_id
+    faber_connection_id = faber_and_alice_connection.faber_connection_id
+
+    did = register_key_issuer
```
ff137, line 264: In cases where `did` is only used once, to assign `credential["ld_credential_detail"]["credential"]["issuer"] = f"{did}"`, we can just replace that with `credential["ld_credential_detail"]["credential"]["issuer"] = register_key_issuer`

app/tests/e2e/issuer/test_issuer_ld_ed25519.py
```diff
@@ -0,0 +1,424 @@
...
+                "degree": {
+                    "type": "BachelorDegree",
+                    "name": "Bachelor of Science and Arts",
+                },
+                "college": "Faber College",
+            },
+            issuanceDate="2021-04-12",
+            issuer="",
+        ),
+        options=LDProofVCDetailOptions(proofType=""),
```
ff137, line 71: All credentials in this module will have `"proofType": "Ed25519Signature2018"`. Note to update the commented example of json too

app/tests/e2e/issuer/test_issuer_ld_ed25519.py
```diff
@@ -0,0 +1,424 @@
...
+    # nothing currently in alice's records
+    assert len(records) == 0
+
+    faber_pub_did = (await faber_acapy_client.wallet.get_public_did()).result.did
+
+    # Updating JSON-LD credential did:sov
+    credential["connection_id"] = faber_connection_id
+    credential["ld_credential_detail"]["credential"][
+        "issuer"
+    ] = f"did:sov:{faber_pub_did}"
```
ff137, line 243: Is using did:sov here intended?

ff137 (changes requested):
Recommended folder structure:
```
e2e		issuer		did_key_bbs		ld_bbs
e2e		issuer		did_key_bbs		ld_x
e2e		issuer		did_key_ed		ld_ed
e2e		issuer		did_key_ed		ld_x
e2e		issuer		did_sov			v1_indy
e2e		issuer		did_sov			v2_indy
```

ff137 (approved):
I noticed that `issuer/did_key_ed/` had tests where did:sov was used for the issuer -- and that there wasn't a test for only did:sov with jsonld - so I added that to `did_sov/test_v2_ld`

PR #445, opened 2023-09-14
Map `problem_report` webhook event

### LINKED ISSUE #446: Add webhook topic mapping / event transformer for `problem-report`

cl0ete:
- Added `problem_report` model.
- Added transformer to webhook service.
- Added `problem_report` topic mapping

Closes #446


ff137 (approved):
Looks good

Do we have a test that generates a problem report? Perhaps we can add to the test an SseListener that awaits it, to ensure the model is mapped properly

PR #479, opened 2023-10-03
:bug: Fix ACA-Py race condition in configuration of endorser connection

ff137:
When onboarding an issuer that does not yet have a public did, a connection first needs to be created between the endorser and the to-be issuer.

Setting up the endorser connection metadata happens asynchronously in ACA-Py, and takes some moments to complete after `set_endorser_role` responds with a success. This leads to some race conditions, where the subsequent `set_endorser_info` method is called too quickly, because the connection metadata has not yet completely updated in ACA-Py storage.

So, it is necessary to assert that roles have been set (in the connection metadata) before the 'info' can be set, and all of this needs to be set before the issuer's did can be registered on the ledger.

This PR adds assert methods in the onboarding logic for an issuer, such that the race condition is eliminated and success rate of onboarding an issuer is more deterministic and reliable.

### DIFF
--- a/app/services/onboarding.py
+++ b/app/services/onboarding.py
@@ -240,17 +240,20 @@ async def onboard_issuer_no_public_did(
         return endorser_connection, connection_record
 
     async def set_endorser_roles(endorser_connection, connection_record):
+        endorser_connection_id = endorser_connection["connection_id"]
+        issuer_connection_id = connection_record.connection_id
+
         bound_logger.debug("Setting roles for endorser")
         await endorser_controller.endorse_transaction.set_endorser_role(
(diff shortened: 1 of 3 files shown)


app/services/onboarding.py
```diff
@@ -148,10 +152,8 @@ async def onboard_issuer(
         bound_logger.debug("No public DID for the to-be issuer")
         # Onboarding an issuer with no public DID can fail when creating a connection with
         # the endorser. If something goes wrong, the whole coroutine should be re-attempted
```
cl0ete, line 154: Nitpick: Update comment

cl0ete (approved):

wdbasson (approved):

PR #491, opened 2023-10-05
Fix coverage report

wdbasson:
Combine coverage reports from matrix steps and upload to Codacy.

### DIFF
--- a/.github/workflows/continuous-deploy.yml
+++ b/.github/workflows/continuous-deploy.yml
@@ -26,7 +26,7 @@ on:
     types: [submitted]
 
 concurrency:
-  group: continuous-deploy-${{ github.ref }}
+  group: cicd-${{ github.ref }}
   cancel-in-progress: true
 
 jobs:
@@ -36,6 +36,10 @@ jobs:
       id-token: write # This is required for requesting the JWT
       packages: write # To push to GHCR.io
     runs-on: ubuntu-latest
+
+    concurrency:
+      group: cicd-build-${{ matrix.image }}}
+      cancel-in-progress: false
 
     outputs:
       image_version: ${{ steps.meta.outputs.version }}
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous-deploy.yml
```diff
@@ -214,10 +218,49 @@ jobs:
               echo "Tests passed with teardown error"
               exit 0
           else
-              exit $EXIT_CODE
+              # exit $EXIT_CODE
```
ff137, line 221: should this be uncommented again?

.github/workflows/continuous-deploy.yml
```diff
@@ -37,6 +37,10 @@ jobs:
       packages: write # To push to GHCR.io
     runs-on: ubuntu-latest
 
+    concurrency:
+      group: cicd-build-${{ matrix.image }}}
+      cancel-in-progress: false
```
ff137, line 42: should this also be true?
or is this covered by the first setting:
```
  group: cicd-${{ github.ref }}
  cancel-in-progress: true
```

.github/workflows/continuous-deploy.yml
```diff
@@ -229,6 +272,9 @@ jobs:
       id-token: write # Required to authenticate with AWS
       contents: read # Required to clone this repository
     runs-on: ubuntu-latest
+    concurrency:
+      group: cicd-deploy
+      cancel-in-progress: false
```
ff137, line 277: similar to cicd-build - shouldn't we cancel in progress deploy jobs if new push comes in?

wdbasson: Not currently no. We don't want to cancel an actual deployment halfway through the process.

.github/workflows/continuous-deploy.yml
wdbasson: No, this overrides the global concurrency config.  We don't want the same images to be built concurrently, it messes with the cache.

ff137: Aah, right. I'm used to the build job including the pytest job :+1:

rblaine95 (approved):

ff137 (approved):

.github/workflows/continuous-deploy.yml
wdbasson: We could configure concurrency per branch, where `cicd-build-${{ github.ref }}-${{ matrix.image }}`, and cancel in progress, but I want to test some stuff first, because the caching is fickle.

.github/workflows/continuous-deploy.yml
wdbasson: Yes. Thanks

PR #498, opened 2023-10-12
Optimize Dockers

rblaine95:
* Don't use official FastAPI docker image
  * According to FastAPI Documentation, you shouldn't use the official
FastAPI docker image in k8s
  * Also add `--no-cache-dir` to `pip install` commands to reduce docker
  image size
* Use `slim` docker images
image


rblaine95:
image

ff137 (approved):
nice!

PR #505, opened 2023-10-16
:boom::sparkles: Use new cloudcontroller and upgrade pydantic to v2

ff137:
Following on from the CloudController upgrade: https://github.com/didx-xyz/aries-cloudcontroller-python/pull/114

The refactored CloudController is mostly a drop-in replacement, as the AcaPyClient still functions the same. The breaking changes are actually quality-of-life improvements, such as allowing us to upgrade to pydantic v2, which amounts to most of the changes in this PR, like `.dict()` -> `.to_dict()` or `.model_dump()` renames; validator changes; etc

Other noteworthy changes:
- custom overrides of the JsonLD API (to circumvent overly strict model validation) and the Multitenancy API (to support the group_id plugin for wallets), these have been moved into the CloudController default APIs
- error handling changes for CloudController's new custom exception
- :boom: some model fields (the ones using reserved words - `from` and `schema`) have been renamed from being appended with an underscore, to instead being prepended with `var_`. Example: 
  ```py
    indy_proof_request = IndyProofRequest(
        non_revoked=IndyProofRequestNonRevoked(var_from=0, to=20),
    ...
    ```
  Json request bodies can still use "from" as a key name - and that is the default. However, the breaking change would be that "from_" is no longer valid. Particularly, this impacts almost all verifier routes request and response models.
- :boom: the `p_value` field in `requested_predicates` for ProofRequest model types, is now strictly an int


app/models/tenants.py
```diff
@@ -12,45 +10,37 @@ class CreateWalletRequestWithGroups(CreateWalletRequest):
...
-class WalletRecordWithGroups(WalletRecord):
-    group_id: Optional[str] = Field(None, example="SomeGroupId")
-
-
-class WalletListWithGroups(BaseModel):
-    results: Optional[List[WalletRecordWithGroups]] = None
-
-
 class TenantRequestBase(BaseModel):
-    image_url: Optional[HttpUrl] = Field(
```
ff137, line 24: todo - set this back to HttpUrl

app/models/tenants.py
```diff
@@ -12,45 +10,37 @@ class CreateWalletRequestWithGroups(CreateWalletRequest):
...
+    group_id: Optional[str] = Field(None, examples=["SomeGroupId"])
 
 
 class Tenant(BaseModel):
-    tenant_id: str = Field(..., example="545135a4-ecbc-4400-8594-bdb74c51c88d")
-    tenant_name: str = Field(..., example="Alice")
-    image_url: Optional[str] = Field(None, example="https://yoma.africa/image.png")
+    tenant_id: str = Field(..., examples=["545135a4-ecbc-4400-8594-bdb74c51c88d"])
+    tenant_name: str = Field(..., examples=["Alice"])
+    image_url: Optional[str] = Field(None, examples=["https://yoma.africa/image.png"])
```
ff137, line 36: Also test setting this to Url type

ff137 (commented):
Also review any remaining pydantic deprecation warnings. Edit: done :white_check_mark:

rblaine95 (approved):
For what my review is worth, LGTM 👍

PR #521, opened 2023-10-25
:white_check_mark: fix spurious test failures with short sleep

ff137:
The fact that our HTTP requests are now much faster (thanks to re-using SSL certs), seems to impact tests that fetch credentials, where a very short delay might be necessary, between awaiting the credential state webhook and fetching the credentials

### DIFF
--- a/app/tests/e2e/issuer/did_key_bbs/test_ld_bbs.py
+++ b/app/tests/e2e/issuer/did_key_bbs/test_ld_bbs.py
@@ -1,3 +1,4 @@
+import asyncio
 from copy import deepcopy
 
 import pytest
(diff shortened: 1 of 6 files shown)


app/tests/e2e/issuer/did_key_bbs/test_ld_bbs.py
```diff
@@ -106,6 +107,7 @@ async def test_send_jsonld_key_bbs(
     )
 
     # Check if Alice received the credential
+    asyncio.sleep(0.1)  # credential may take moment to reflect after webhook
```
ff137, line 110: It would've been nice if I added `await` in front of the coroutine, but it's the thought that counts :see_no_evil:

PR #527, opened 2023-10-31
Status check hotfix

wdbasson:
Check if any test failed and the fail `Status Check` job. Can be used as a single status check to verify that tests have passed.

### DIFF
--- a/.github/workflows/continuous-deploy.yml
+++ b/.github/workflows/continuous-deploy.yml
@@ -234,9 +234,9 @@ jobs:
               exit 0
           else
               if [ "$EXIT_CODE" -ne 0 ]; then
-                echo "::set-output name=test_success::false"
+                echo "test_success=false" >> $GITHUB_OUTPUT
               else
-                echo "::set-output name=test_success::true"
+                echo "test_success=true" >> $GITHUB_OUTPUT
               fi
               exit $EXIT_CODE
           fi
@@ -258,8 +258,8 @@ jobs:
 
   combine-coverage:
     name: Coverage
-    needs: status-check
     runs-on: ubuntu-latest
+    needs: status-check
     steps:
       - uses: actions/checkout@v4
         with:
--- a/.github/workflows/continuous-deploy.yml
+++ b/.github/workflows/continuous-deploy.yml
@@ -153,6 +153,8 @@ jobs:
     needs:
       - build
     runs-on: ubuntu-latest
+    outputs:
+      test_success: ${{ steps.test.outputs.test_success }}
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


.github/workflows/continuous-deploy.yml
```diff
@@ -231,6 +233,11 @@ jobs:
               echo "Tests passed with teardown error"
               exit 0
           else
+              if [ "$EXIT_CODE" -ne 0 ]; then
+                echo "::set-output name=test_success::false"
+              else
+                echo "::set-output name=test_success::true"
```
rblaine95, line 239: > The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

rblaine95 (changes requested):

ff137 (approved):

.github/workflows/continuous-deploy.yml
```diff
@@ -239,10 +246,20 @@ jobs:
           name: coverage-files-${{ matrix.module.id }}
           path: coverage-files/.coverage.${{ matrix.module.id }}
 
+  status-check:
```
rblaine95, line 249: As far as I remember, if any step within a matrix fails (by default) then all steps that have not yet completed are automatically canceled and all subsequent steps are skipped. 

Is this status check step required?

.github/workflows/continuous-deploy.yml
```diff
@@ -231,6 +233,11 @@ jobs:
               echo "Tests passed with teardown error"
               exit 0
           else
+              if [ "$EXIT_CODE" -ne 0 ]; then
+                echo "test_success=false" >> $GITHUB_OUTPUT
+              else
+                echo "test_success=true" >> $GITHUB_OUTPUT
+              fi
               exit $EXIT_CODE
```
rblaine95, line 241: This `exit $EXIT_CODE` would exit with a non-zero exit code, marking the github step as having failed, which would cancel subsequent jobs from running.

Is the `outputs.test_success` necessary? Wouldn't the job failing handle it?

.github/workflows/continuous-deploy.yml
```diff
@@ -239,10 +246,20 @@ jobs:
...
+    steps:
+      - name: Check if any test failed
+        run: exit 1
+        if: needs.test.outputs.test_success == 'false'
+
   combine-coverage:
     name: Coverage
-    needs: test
     runs-on: ubuntu-latest
+    needs: status-check
```
rblaine95, line 262: If any job in the test matrix fails, then the `combine-coverage` job would have been skipped.

Wouldn't it be easier to add `Coverage` as a required check to PR's instead of adding `Status Check` and the above `if else`?

PR #551, opened 2023-11-22
:sparkles: Add caching to schema_id fetch from ledger

ff137:
Note: `beaker`, the cache library, has a critical vulnerability:
https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2013-7489
An insecure data serialization method could theoretically allow arbitrary code execution.


ff137:
Beaker has a critical vulnerability, and it seems like caching won't improve scalability significantly enough, so closing for now

PR #635, opened 2024-01-31
:construction: Backlog: Add `aries-acapy-cache-redis` plugin

ff137:
:heavy_plus_sign: Adds [aries-acapy-cache-redis plugin](https://github.com/Indicio-tech/aries-acapy-cache-redis/tree/main) to the multitenant and governance ACA-Py agent.
:construction_worker: Configures an extra redis instance to use for caching

Edit: moving to the backlog, as we should do some before-after load testing to validate if these makes any improvement. First thought it would solve HA problems, but it's no longer needed


ff137:
OBE

PR #664, opened 2024-02-27
Add query params to get connections and credential/proof exchange records 

### LINKED ISSUE #293: Implement /generic/verifier/proofs/{connection_id} endpoint in cloud api
Good day @TimoGlastra ,

We have a requirement on the aries-cloud-api to have the following endpoint available so that we are able to retrieve proofs per connection_id.
/generic/verifier/proofs/{connection_id}

Hope to see the functionality available soon as we need it to display certain information regarding a connection.

Groetjes

Jurgens

cl0ete:
Added query params to:
- Get connections
- Get credentials exchange records
- Get proof exchange records

Updated models where needed
Updated conversion map with missing states.
Add dict to revert state back to v1 state in order to  query records
Updated doc strings

Updated postgres to 16 (@rblaine95  request) 
closes #293


shared/models/conversion.py
```diff
@@ -20,6 +21,8 @@
     PresentationExchange,
 )
 
+logger = get_logger(__name__)
```
ff137, line 24: Ah, because the `__name__` for this module is in shared, it will log to what appears to be a new service, called shared. Same if file logging was activated, it would write to a directory called shared. 

So, the recommendation is to pass the logger from wherever these functions are called. i.e.  
```py
from logging import Logger
...
def v1_presentation_state_to_rfc_state(state: Optional[str], logger: Logger) -> Optional[str]:
```

And then add Logger to: `presentation_record_to_model(.., logger: Logger)`

Then, wherever that is called in webhooks, or the verifier facades, pass the relevant logger from there

app/tests/e2e/test_connections.py
```diff
@@ -109,6 +109,55 @@ async def test_get_connections(
...
+        await alice_member_client.get(
+            f"{BASE_PATH}?their_public_did={bob_connections[0]['their_public_did']}"
+        )
+    assert exc.value.status_code == 422
+
+    alice_their_role = (
+        await alice_member_client.get(f"{BASE_PATH}?their_role=inviter")
+    ).json()[0]["their_role"]
+    assert alice_their_role == "inviter"
+
```
ff137, line 160: Just apply a lil bit of deduplication here: namely with all the `bob_connections[0]` and `alice_connections[0]` calls. The [0] can be dropped just by re-assigning, e.g., `alice_connection = alice_connections[0]`. And any other fields that are duplicated, like `alice_connections[0]["my_did"]` and `["invitation_msg_id"]`

app/tests/issuer/test_issuer.py
```diff
@@ -100,28 +100,49 @@ async def test_get_credentials(
     with when(IssuerV1).get_records(...).thenReturn(
         to_async(v1_records_no_conn_id)
     ), when(IssuerV2).get_records(...).thenReturn(to_async(v2_records_no_conn_id)):
-        result = await test_module.get_credentials(None, mock_tenant_auth)
+        result = await test_module.get_credentials(
+            None, None, None, None, mock_tenant_auth
```
ff137, line 104: best to name these args, to know what these Nones are assigning to

app/routes/issuer.py
```diff
@@ -35,28 +39,51 @@
 
 @router.get("", response_model=List[CredentialExchange])
 async def get_credentials(
-    connection_id: Optional[str] = Query(None),
+    connection_id: Optional[UUID] = Query(None),
```
ff137, line 42: Is it correct to use UUID type here? Are all connection-ids of UUID type? And does it just add some extra validation? If so, may as well implement it everywhere! Because we currently mostly use str type for connection_id. So either str everywhere, or UUID everywhere.

ff137 (commented):
Looks good!

ff137 (approved):
Good stuff!
I just made one change of moving the Literal definitions (State, Role, Protocol) to the shared models. So that they can also use the Literals in their state/role types

PR #688, opened 2024-03-13
Update proof models

cl0ete:
Extend `IndyProofRequest` model add new default values for `name` and `version`
Catch ApiException on accept proof request

Test proof request with no version or name fields 
mark with pytest xfail (expect test to fail) raise `HTTPException strict = True`

### DIFF
--- a/app/services/verifier/acapy_verifier_v1.py
+++ b/app/services/verifier/acapy_verifier_v1.py
@@ -1,6 +1,5 @@
 from aries_cloudcontroller import (
     AcaPyClient,
-    ApiException,
     V10PresentationCreateRequestRequest,
     V10PresentationProblemReportRequest,
     V10PresentationSendRequest,
@@ -135,18 +134,11 @@ class VerifierV1(Verifier):
                 body=v10_pres_send_req,
             )
             result = record_to_model(presentation_record)
-        except ApiException as e:
-            bound_logger.exception(
-                "An ACA-PY ApiException occurred while sending a proof presentation."
-            )
-            raise CloudApiException(
-                f"Failed to send proof presentation: {e.reason}."
-            ) from e
         except Exception as e:
             bound_logger.exception(
-                "An unexpected error occurred while sending a proof presentation."
+                "An exception occurred while sending a proof presentation."
             )
-            raise CloudApiException("Failed to send proof presentation.") from e
+            raise CloudApiException(f"Failed to send proof presentation: {e}.") from e
 
         if result:
             bound_logger.debug("Successfully sent v1 proof presentation.")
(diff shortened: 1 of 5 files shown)


app/services/verifier/acapy_verifier_v1.py
```diff
@@ -1,5 +1,6 @@
 from aries_cloudcontroller import (
     AcaPyClient,
+    ApiException,
```
ff137, line 3: We can revert these changes (and in _v2 service as well). Error handling already improved in incoming PR

app/tests/e2e/test_proof_request_models.py
```diff
@@ -0,0 +1,121 @@
...
+from app.routes.verifier import router as verifier_router
+from app.services.event_handling.sse_listener import SseListener
+from app.tests.util.ecosystem_connections import AcmeAliceConnect
+from shared.util.rich_async_client import RichAsyncClient
+
+VERIFIER_BASE_PATH = verifier_router.prefix
+
+
+@pytest.mark.anyio
+@pytest.mark.xfail(raises=HTTPException, strict=True)
```
ff137, line 20: I think xfail takes a "reason" argument, which we should populate. Something like: 
"Assert `version` and `name` are required fields, despite being optional according to ACA-Py models".

app/tests/e2e/test_proof_request_models.py
```diff
@@ -0,0 +1,121 @@
...
+                        "THE_SPEED": {
+                            "name": "speed",
+                            "restrictions": [],
+                        }
+                    },
+                    "requested_predicates": {},
+                },
+                auto_verify=True,
+            )
+        )
```
ff137, line 65: nitpick: we should get in the habit of defining models outside of the method argument. So:
```
request_body=V10PresentationSendRequestRequest(
                auto_remove=False,
                connection_id=acme_connection_id,
                comment="Test proof",
                proof_request={
                    "name": name,
                    "version": version,
                    "requested_attributes": {
                        "THE_SPEED": {
                            "name": "speed",
                            "restrictions": [],
                        }
                    },
                    "requested_predicates": {},
                },
                auto_verify=True,
            )

await acme_acapy_client.present_proof_v1_0.send_request_free(body=request_body)
```

Easier to read, and to debug cases where model creation fails vs method fails.

Same with method in the else condition:

app/tests/e2e/test_proof_request_models.py
```diff
@@ -0,0 +1,121 @@
...
+
+    # Get referent
+    referent = (
+        await alice_member_client.get(
+            f"{VERIFIER_BASE_PATH}/proofs/{alice_proof_exchange_id}/credentials"
+        )
+    ).json()[0]["cred_info"]["referent"]
+
+    # Send proof
+    await alice_member_client.post(
```
ff137, line 107: Maybe just a comment here indicating that this is what will raise the expected exception

ff137 (changes requested):
Nice! Just some minor changes for the tests, and then please revert the acapy_verifier_v1 / v2 changes (will cause merge conflicts with incoming exception handling improvements)

ff137 (approved):
Nice!

PR #693, opened 2024-03-15
:construction: test issuing a credential on deleted connection

ff137:


ff137:
Not resolvable with deprecated Connections API. A solution to be tested with new DidRotate API

PR #707, opened 2024-03-25
:sparkles: obfuscate secret data from webhook logs

ff137:
Two webhook event topics were identified to possibly contain sensitive data:
`endorsement` events, and `issue_credential_v2_0_indy` events.

Endorsements can either contain a `messages_attach`, or `signature_response` object in the payload, which contains a json string that references a `master_secret` value. The exact nature of this value, and whether it is truly sensitive data, is not yet fully known, but for simplicity we redact these values from logs.

Similarly, `issue_credential_v2_0_indy` events contain a `master_secret_blinding_data` field in the `cred_request_metadata`. For simplicity, we redact these values as well.

Lastly, loguru sinks take a `diagnose` field, which impacts whether values are printed in exception stack traces or not. To prevent sensitive data from leaking in exceptions, we set this to be False by default, for serialised logging only.

### DIFF
--- a/app/util/retry_method.py
+++ b/app/util/retry_method.py
@@ -77,7 +77,7 @@ async def coroutine_with_retry_until_value(
                     return result
 
             if attempt + 1 < max_attempts:
-                logger.info(
+                logger.debug(
                     (
                         "Coroutine returned {} instead of expected {} "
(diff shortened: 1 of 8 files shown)


cl0ete (approved):
LGTM :+1:

PR #712, opened 2024-03-26
:sparkles: Validate wallet_id belongs to group_id in SSE subscriptions

ff137:
Adds assertion logic to SSE routes, so that if a group_id exists in the query parameters. we assert using redis keys that the wallet does in fact belong to that group. We raise a 403 otherwise.

### DIFF
--- a/webhooks/services/webhooks_redis_serivce.py
+++ b/webhooks/services/webhooks_redis_serivce.py
@@ -332,6 +332,32 @@ class WebhooksRedisService(RedisService):
         if result:
             bound_logger.debug("Successfully wrote endorsement entry to redis.")
         else:
-            bound_logger.warning(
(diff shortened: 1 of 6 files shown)


webhooks/services/webhooks_redis_serivce.py
```diff
@@ -319,6 +319,32 @@ def add_endorsement_event(self, event_json: str, transaction_id: str) -> None:
         if result:
             bound_logger.debug("Successfully wrote endorsement entry to redis.")
         else:
-            bound_logger.warning(
-                "Redis key for this endorsement entry is already set."
-            )  #
```
ff137, line 324: Not sure what happened here. Accidentally remove a couple lines in a past commit!

cl0ete (approved):

PR #715, opened 2024-03-27
Add Lago billing manager

cl0ete:
Add billing manager .
If no LAGO_API_KEY is given events will not be processed for billing.

Write billing events to redis with `group_id` as redis key. 

Billing on:

- Ledger writes
- Proofs
- Credentials issued
- Credentials revoked


webhooks/services/acapy_events_processor.py
```diff
@@ -338,6 +340,38 @@ def _process_event(self, event_json: str) -> None:
...
+                topic=cloudapi_topic, payload=payload, logger=bound_logger
+            )
+        ):
+            bound_logger.info(
+                "Forwarding billing event for Billing service",
+            )
+            if cloudapi_topic == "endorsements":
+                # Add the operation type to the payload for endorsements
+                # Simplifies the billing service's logic for determining the operation type
+                endorse_event: Dict[str, Any] = json.loads(webhook_event_json)
```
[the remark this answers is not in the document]
ff137: But, we should avoid serialising / deserialising if possible. 

You're essentially modifying the payload to include the "type" key with the operation_type (which I presume is because operation type is nested in a json string in message_attach: data, etc. 
We can do that without loading the json.

Earlier we had:
```
webhook_event_json = cloudapi_webhook_event.model_dump_json()
```

You can copy the `cloudapi_webhook_event`, replace payload directly (because already have the payload dict in scope). So, payload["type"] = operation_type, which is fine because payload is not used elsewhere again. Then `modified_model = cloudapi_webhook_event.copy(payload = payload)` <- something like that, Then 
`webhook_event_for_billing = modified_model.model_dump_json()` <- use pydantic's method to do it.

So, we then just have 1 more json operation, instead of 2

Then to deduplicate the .add_billing_event call (like mentioned in other comment), you can just say:
```
            if cloudapi_topic == "endorsements":
...
                webhook_event_for_billing = modified_model.model_dump_json()

            else:
                webhook_event_for_billing = webhook_event_json

            self.redis_service.add_billing_event(
                event_json=webhook_event_for_billing ,
                group_id=group_id,
                wallet_id=wallet_id,
                timestamp_ns=event.metadata.time_ns,
            )

File: webhooks/services/billing_processor.py [hunk 2]
```diff
@@ -0,0 +1,275 @@
...
+
+    async def _process_billing_event(self, message: Dict[str, Any]) -> None:
+        """
+        Process billing events. Convert them to LAGO events and post them to LAGO
+        """
+        message_data = message.get("data")
+        if isinstance(message_data, bytes):
+            message_data = message_data.decode("utf-8")
+
+        groub_id, timestamp_ns_str = message_data.split(":")
```
ff137, line 134: group* :-)

webhooks/services/billing_processor.py
```diff
@@ -0,0 +1,275 @@
...
+
+        if not events:
+            # there are duplicate done event bc of acapy to cloudapi conversion
+            # the score get updated and the event is not found with old score
+            logger.debug(
+                f"No events found for group_id: {groub_id} and timestamp: {timestamp_ns}"
+            )
+            return
+
+        event: Dict[str, Any] = json.loads(events[0])
```
ff137, line 149: orjson

webhooks/services/billing_processor.py
```diff
@@ -0,0 +1,275 @@
...
+        if event.get("topic") == "endorsements":
+            # handel endorsements event
+            lago = self._convert_endorsements_event(event)
+
+        if event.get("topic") == "issuer_cred_rev":
+            # handel issuer_cred_rev event
+            lago = self._convert_issuer_cred_rev_event(event)
+
+        # post billing event to LAGO
+        await self._post_billing_event(lago)
```
ff137, line 168: lago could be undefined if no ifs matched. I'd say use a if-else tree (can be replaced with switch in python 3.12), and then last else just logs a warning (unexpected topic) and returns

webhooks/services/billing_processor.py
```diff
@@ -0,0 +1,275 @@
...
+
+    async def _post_billing_event(self, event: BillingEvent) -> None:
+        """
+        Post billing event to LAGO
+        """
+        logger.debug(f"Posting billing event: {event}")
```
ff137, line 174: just check all the logs not to use f-interpolation. Use the `logger.debug("Posting billing event: {}", event)` pattern instead. Because that skips the string operation if log level is not said to debug. using f-string will do the operation even if log level is not debug.

Just search everywhere for "logger.debug(f" (and info, error etc) to check any other logs that might have this,

webhooks/services/billing_processor.py
```diff
@@ -0,0 +1,275 @@
...
+
+    def __init__(self, redis_service: WebhooksRedisService) -> None:
+        self.redis_service = redis_service
+
+        self._tasks: List[asyncio.Task] = []
+
+        self._pubsub = None
+
+        self._client = RichAsyncClient(name="BillingManager")
+        self._client.headers = {"Authorization": f"Bearer {LAGO_API_KEY}"}
```
ff137, line 39: I think can be defined in 1 line. Pass headers as kwarg to RichAsyncClient

webhooks/services/billing_processor.py
```diff
@@ -0,0 +1,275 @@
...
+        """
+        logger.debug(f"Posting billing event: {event}")
+        try:
+            lago_response = await self._client.post(
+                url=self._LAGO_URL,
+                json={"event": event.model_dump()},
+            )
+
+            logger.error(f"Response from LAGO: {lago_response.json()}")
+            lago_response.raise_for_status()
```
ff137, line 181: Ah ... btw, RichAsyncClient does this automatically. That's part of what makes it a RichAsyncClient ;-) it does the exception raising for us.

So that's why I added a ```
        raise_status_error=True,
``` 
flag to the rich async client. To turn off automatically raising, such that error handling can be done differently if desired.

In this case I guess you can just skip the raise.

File: webhooks/services/billing_processor.py [hunk 8]
```diff
@@ -0,0 +1,275 @@
...
+
+        except HTTPException as e:
+            if e.status_code == 422:
+                logger.error(f"Error posting billing event >>> : {e.detail}")
+                pass
+            logger.error(f"Error posting billing event: {e}")
+            raise e
```
ff137: Other status codes will raise an exception which should be handled where this method is called ... otherwise it'll kill the async loop (if this is running in a loop). Just make sure the exception is caught at some level

webhooks/services/billing_processor.py
```diff
@@ -0,0 +1,275 @@
...
+        """
+        Convert credential event to LAGO event
+        """
+        logger.debug(f"Converting credential event: {event}")
+
+        # using thread_id as transaction_id
+        payload: Dict[str, Any] = event.get("payload")
+        lago_event = CredentialBillingEvent(
+            transaction_id=payload.get("thread_id"),
+            external_customer_id=event.get("group_id"),
```
ff137, line 204: Keep in mind, these CAN be None ... If Lago handles null values, it's fine. Otherwise we might want to add `event.get("group_id") or "no group"` or something so it has a default empty value. Not sure if it's necessary

webhooks/services/billing_processor.py
```diff
@@ -0,0 +1,275 @@
...
+                external_customer_id=event.get("group_id"),
+            )
+            return lago_event
+
+        if payload.get("type") == "113":
+            lago_event = RevRegDefBillingEvent(
+                transaction_id=payload.get("transaction_id"),
+                external_customer_id=event.get("group_id"),
+            )
+            return lago_event
```
ff137: Veeeery similar method operations happening here. The only difference is the class being constructed. We are defining `payload.get("transaction_id")` each time. Rather just do that once at the beginning. Or maybe GPT has a more elegant way to define classes dynamically which have the same attributes.

Actually, it makes me wonder why different classes is necessary ... just because they have the Literal attribute? Maybe we should just define that dynamically here, and then there's one Class being constructed, but with one of many codes. Idk, might be more readable that way.

webhooks/services/billing_processor.py
```diff
@@ -0,0 +1,275 @@
...
+        Convert issuer cred rev event to LAGO event
+        """
+
+        # using record_id as transaction_id
+        logger.debug(f"Converting issuer cred rev event: {event}")
+        payload: Dict[str, Any] = event.get("payload")
+        lago_event = RevocationBillingEvent(
+            transaction_id=payload.get("record_id"),
+            external_customer_id=event.get("group_id"),
+        )
```
ff137, line 278: Similarly, each method is unpacking payload in the same way. Just pass the payload to these functions, instead of event.

webhooks/services/billing_processor.py
```diff
@@ -0,0 +1,275 @@
...
+
+        if event.get("topic") == "proofs":
+            # handel proofs event
+            lago = self._convert_proofs_event(event)
+
+        if event.get("topic") == "endorsements":
+            # handel endorsements event
+            lago = self._convert_endorsements_event(event)
+
+        if event.get("topic") == "issuer_cred_rev":
```
ff137: What might be better is to say topic = event.get("topic"); if topic == "issuer_cred_rev", then transaction_id=payload.get("record_id"), and `code = "revoked"`. So each model is a billable event, where they're constructed in the same way. They take the same payload: a transaction_id, a group_id, and a code.

webhooks/util/billing.py
```diff
@@ -0,0 +1,41 @@
...
+        "presentation_acked",
+    ]:
+        logger.debug(f"Event state: {state} is not applicable for the billing service.")
+        return False
+
+    if topic == "endorsements":
+        operation_type = get_operation_type(payload=payload, logger=logger)
+        if operation_type not in ["1", "100", "102", "113", "114"]:
+            logger.debug(
+                f"Endorsement operation type: {operation_type} is not applicable for the billing service."
```
ff137, line 29: Again, just check all debug logs for f-interpolation please. This one won't be caught with a `logger.debug(f` global search :-) because of the \n

webhooks/services/billing_processor.py
```diff
@@ -0,0 +1,301 @@
...
+
+        elif topic == "issuer_cred_rev":
+            # handel issuer_cred_rev event
+            record_id = payload.get("record_id")
+            lago = self._convert_issuer_cred_rev_event(
+                group_id=group_id, record_id=record_id
+            )
+
+        else:
+            logger.warning("Unknown billing event: {}", event)
```
ff137, line 190: unknown topic

webhooks/services/billing_processor.py
```diff
@@ -0,0 +1,301 @@
...
+            logger.error("Pubsub is not running")
+
+        all_running = self._tasks and all(not task.done() for task in self._tasks)
+        logger.debug("All tasks running: {}", all_running)
+
+        if not all_running:
+            for task in self._tasks:
+                if task.done():
+                    logger.warning("Task `{}` is not running", task.get_name())
+        return self._pubsub and all_running
```
ff137, line 83: I wonder if this will cause problems with the health check when billing processor isn't defined.

By default, `are_task_running` (typo btw -> tasks) -- by default, self._pubsub will be None, and all_running could also be none because there are no self._tasks.

The reason I'm pointing this out is because we don't have test coverage for the healthcheck endpoints (they need dependency overriding that I haven't figured out yet). So this could break the healthcheck without us knowing from tests. So just keep in mind that if Lago doesn't exist, then the BillingProcessor should not start up, and the "are_tasks_running" check which is called by the healthcheck should return TRUE when Lago is undefined and no tasks exist ...

webhooks/services/acapy_events_processor.py
```diff
@@ -338,6 +341,39 @@ def _process_event(self, event_json: str) -> None:
...
+                bound_logger.debug(
+                    "Forwarding billing event for Billing service",
+                )
+                if cloudapi_topic == "endorsements":
+                    # Add the operation type to the payload for endorsements
+                    # Simplifies the billing service's logic for determining the operation type
+
+                    operation_type = get_operation_type(
+                        payload=payload, logger=bound_logger
+                    )
```
ff137, line 361: Note: `get_operation_type` is effectively called twice -- once in `is_applicable_for_billing`, and once again here ...

One possible optimisation is to make is_applicable_for_billing return a tuple ... i.e. the True/False result, but also the operation_type ... so we say:

```
if LAGO_API_KEY:
    is_applicable, operation_type = is_applicable_for_billing(
                wallet_id=wallet_id,
                group_id=group_id,
                topic=cloudapi_topic,
                payload=payload,
                logger=bound_logger,
            )
  if is_applicable:
    
    if cloudapi_topic == "endorsements":
      # we already have operation_type returned from the `is_applicable_for_billing` check, so just add to [payload][type]
```

Optimisation is the root of all evil, so just an idea! Because get_operation_type does a bunch of key extraction + json transforms

webhooks/models/billing_payloads.py
```diff
@@ -0,0 +1,47 @@
...
+    REVOCATION = "revoked"
+
+
+class LagoEvent(BaseModel):
+    transaction_id: str
+    external_customer_id: str
+
+
+class CredentialBillingEvent(LagoEvent):
+    code: Literal[LagoTopics.CREDENTIAL] = Field("issue_done")
```
ff137, line 23: Ah, the codes in the Field() should probably use LagoTopics.CREDENTIAL too. Otherwise `issue_done` is duplicated

webhooks/services/billing_processor.py
```diff
@@ -0,0 +1,301 @@
...
+                external_customer_id=group_id,
+            )
+            return lago_event
+
+        elif endorsement_type == TransactionTypes.REVOC_REG_ENTRY:
+            lago_event = RevRegEntryBillingEvent(
+                transaction_id=transaction_id,
+                external_customer_id=group_id,
+            )
+            return lago_event
```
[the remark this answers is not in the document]
cl0ete: I like this :+1:

webhooks/services/billing_processor.py
cl0ete: At the moment all I am doing is not passing events on to the service if there is no `apikey` so the service is running... right?

webhooks/services/billing_processor.py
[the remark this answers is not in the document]
cl0ete: Check if you like what i did...

ff137 (changes requested):
Ah, one thing that needs to be added ... test coverage for this util method:
https://github.com/didx-xyz/aries-cloudapi-python/blob/e0ad6cb1ed343f569d812e450c440bb924514928/webhooks/util/billing.py#L17-L57

ff137 (approved):

PR #716, opened 2024-03-28
:sparkles: Configure SseListeners in tests with retry logic

ff137:
Replaces usages of SseListeners in tests with the helper "check_webhook_state" method. 

Implemented some basic retry logic in this helper method. Currently configured to retry on HTTP disconnects once.

### DIFF
--- a/app/tests/util/webhooks.py
+++ b/app/tests/util/webhooks.py
@@ -30,6 +30,8 @@ async def check_webhook_state(
     filter_map: Optional[Dict[str, str]] = None,
     max_duration: int = 60,
     lookback_time: int = 1,
+    max_tries: int = 2,
+    delay: float = 0.5,
 ) -> Dict[str, Any]:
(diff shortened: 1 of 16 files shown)


app/tests/util/webhooks.py
```diff
@@ -20,42 +26,75 @@ def get_wallet_id_from_async_client(client: RichAsyncClient) -> str:
...
-    desired_state = filter_map["state"]
+    # Retry logic in case of disconnect errors (don't retry on timeout errors)
+    event = None
+    attempt = 0
+    max_tries = 2
+    delay = 0.5
 
-    other_keys = dict(filter_map)
-    other_keys.pop("state", None)
+    while not event and attempt < max_tries:
```
ff137, line 47: @cl0ete @wdbasson Here is the retry logic implementation.

Currently tries max 2 times. i.e. just one retry, and only if there's an HTTP Error. Timeouts won't retry.

cl0ete (approved):

PR #735, opened 2024-04-05
:adhesive_bandage: Fix WebSocket callback function

ff137:
`callback` requires the keyword: topic, which was removed as it's unused in the scope of the function. Whoops :see_no_evil:


ff137:
GitHub is experiencing some 500 server issues at the moment. Jobs failing unnecessarily.

I'm just going to merge because this PR makes no changes to anything that's tested

PR #754, opened 2024-04-11
Add nats jetstream cluster for local dev with persistent storage

henrymsiska:
➕ Adds [NATS Jetstream](https://docs.nats.io/nats-concepts/jetstream) cluster to docker compose for local dev to support durable event streaming

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -581,11 +581,63 @@ services:
       governance-multitenant-web:
         condition: service_started
 
+  nats-1:
+    image: nats:alpine
+    ports:
+      - "8222:8222"
+      - "4222:4222"
+    command: "-c /etc/nats/nats.conf --server_name nats-1"
+    volumes:
+      - nats-data-1:/nats-data
+      - ./resources/nats-local.conf:/etc/nats/nats.conf
+    networks:
+      - nats
+    labels:
+      app: nats
+      tier: infra
+    healthcheck:
+      test: ["CMD-SHELL", 'wget -qO- localhost:8222/healthz | grep ''"status":"ok"'' || exit 1']
+      start_period: 30s
+      interval: 10s
+      timeout: 10s
+      retries: 5
+  nats-2:
+    image: nats
+    command: "-c /etc/nats/nats.conf --server_name nats-2"
+    volumes:
+      - nats-data-2:/nats-data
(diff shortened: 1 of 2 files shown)


docker-compose.yaml
```diff
@@ -581,11 +581,63 @@ services:
...
+  nats-3:
+    image: nats
+    command: "-c /etc/nats/nats.conf --server_name nats-3"
+    volumes:
+      - nats-data-3:/nats-data
+      - ./resources/nats-local.conf:/etc/nats/nats.conf
+    networks:
+      - nats
+    depends_on:
+      nats-1:
```
rblaine95, line 627: I think maybe this?
```suggestion
      nats-2:
```

So `nats-2` depends on `nats-1` which depends on `nats-0`

docker-compose.yaml
```diff
@@ -581,11 +581,63 @@ services:
...
+      app: nats
+      tier: infra
+  nats-3:
+    image: nats
+    command: "-c /etc/nats/nats.conf --server_name nats-3"
+    volumes:
+      - nats-data-3:/nats-data
+      - ./resources/nats-local.conf:/etc/nats/nats.conf
+    networks:
+      - nats
```
rblaine95, line 625: We should make sure that the NATS servers are on the same Compose network as whatever needs to talk to it 🤔

rblaine95 (changes requested):

docker-compose.yaml
henrymsiska: They don't have to chain. The nats-1 dependency is because the first instance is whats exposed and and used from a client perspective so the dependency is just to help ensure that it startsup first. Appart from that, the dependency is not needed.

docker-compose.yaml
henrymsiska: Yes, will add to the relevant ones when we integrate

ff137: It will be the webhooks service: `governance-webhook-web`

ff137: @henrymsiska @rblaine95 any other relevant changes to include? Can we approve / merge?

henrymsiska: @ff137 @rblaine95 : Will add the network to the correct component that will pull from redis to nats once that is added.

rblaine95 (approved):

ff137 (approved):

PR #766, opened 2024-04-17
Add tests for `trustregistry/main.py`

cl0ete:
Thanks ChatGPT for the help

### DIFF
--- a/trustregistry/tests/test_main.py
+++ b/trustregistry/tests/test_main.py
@@ -44,7 +44,7 @@ async def test_lifespan(mock_engine, mock_inspect):
 
 
 @pytest.mark.anyio
-async def test_root():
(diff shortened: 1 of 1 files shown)


trustregistry/tests/test_main.py
```diff
@@ -0,0 +1,76 @@
...
+from sqlalchemy.orm import Session
+
+from trustregistry import db
+from trustregistry.main import app, lifespan, registry, root
+
+
+@pytest.fixture
+def db_session_mock():
+    session = Mock(spec=Session)
+    return session
```
ff137, line 14: cov report shows these lines are uncovered. I see db_session_mock is indeed used, but not as a fixture argument in the tests

PR #779, opened 2024-04-25
 ✨ publishing pending revocations should wait and assert transaction has been acked

ff137:
Required in the 0.12 upgrade -- I figured it's best to implement this in its own PR

Seems to be a timing factor in 0.12, which causes revoked credentials to still appear to be valid in our tests

A fix for that is that the publish revocations endpoint should wait for the ledger transaction to be acked by the endorser before returning a success

To-do later:
- it may be worth including some enriched response instead of a 204.
- We could respond with the number of credentials revoked for this request.

### DIFF
--- a/app/routes/issuer.py
+++ b/app/routes/issuer.py
@@ -1,3 +1,4 @@
+import asyncio
 from typing import List, Optional
 from uuid import UUID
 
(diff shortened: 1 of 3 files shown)


ff137:
Included in #758

PR #802, opened 2024-05-13
:art: rename `credential_id` to `credential_exchange_id` and :warning: mark old field as deprecated

ff137:
The legacy design decision to refer to credential _exchange_ records by "credential_id" is the source of much unnecessary confusion. For sake of clarity, "credential_id" should only be the reference when talking about credentials, and "credential_exchange_id" is the reference for exchange records.

The old attribute name is kept as it is, to allow for users to adapt to the new naming convention before it's removed.


cl0ete (approved):
GG WP

PR #804, opened 2024-05-15
Swagger updates message oob

cl0ete:
Update message and OOB endpoint docstrings


app/routes/messaging.py
```diff
@@ -19,15 +19,21 @@ async def send_messages(
 ):
     """
     Send basic message.
```
ff137, line 21: Just a pedantic request, for consistency -- headings don't need to end in punctuation

app/routes/messaging.py
```diff
@@ -48,15 +54,24 @@ async def send_trust_ping(
 ):
     """
     Trust ping
```
ff137, line 56: This one and previous one can use a custom summary

And for "trust ping", I'm wondering what is a trust ping? Is it just pinging a connection to indicate it's still active?

app/routes/oob.py
```diff
@@ -74,7 +102,27 @@ async def accept_oob_invitation(
     auth: AcaPyAuth = Depends(acapy_auth_from_header),
 ) -> OobRecord:
     """
-    Receive out-of-band invitation.
+    Accept out-of-band invitation.
+    ------------------------------
+
+    As with the connection endpoint, the invitation object from the create-invitation endpoint
```
ff137, line 108: As with the _accept connection invitation_ endpoint

app/routes/oob.py
```diff
@@ -99,19 +147,19 @@ async def connect_to_public_did(
...
+    -------------
+        body: ConnectToPublicDid
+            public_did: str
 
     Returns:
     ---
-    Connection
-        The connection record
+        Connection
+            The connection exchange record
```
ff137, line 162: I think we just call them connection records

app/routes/oob.py
```diff
@@ -22,7 +22,35 @@ async def create_oob_invitation(
...
+
+    ```multi_use``` can't be set to `true` if an ```attachment``` is provided,
+    as a proof request or credential offer should be sent to one tenant.
+
+    Request body:
+    -------------
+        body:CreateOobInvitation
+            alias: Optional[str]
+            multi_use: Optional[bool]
+            use_public_did: Optional[bool]
```
ff137, line 45: I think we should replace all cases of "Optional[bool]" with just bool and a default value, so it's still non-required but the default behaviour is more clear.

Optional bools are always a bit weird, because it's essentially adding a 3rd option to a true / false, and that only makes sense if it has some behavior that's distinct from just plain true / false

app/routes/oob.py
```diff
@@ -22,7 +22,35 @@ async def create_oob_invitation(
...
+            attachments: Optional[List[Attachment]]
+            handshake_protocols: Optional[List[str]]
```
ff137, line 47: I think this was removed from the model

app/routes/oob.py
```diff
@@ -22,7 +22,35 @@ async def create_oob_invitation(
     auth: AcaPyAuth = Depends(acapy_auth_from_header),
 ) -> InvitationRecord:
     """
-    Create connection invitation out-of-band.
+    Create an out-of-band invitation.
+    ------------------------------
+
+    The attachment field is used to include a credential offer or a proof request in the invitation.
+    The attachment ID is the credential exchange ID or proof request ID.
+    The attachment type is either ```"credential-offer"``` or ```"present-proof"```.
```
ff137, line 30: I think the triple ` can just be single ones

ff137 (changes requested):

ff137:
P.S. Let's add RFC links where possible

PR #806, opened 2024-05-16
Add regression test logic to pipeline

wdbasson:


.github/workflows/continuous-deploy.yml
```diff
@@ -620,30 +715,97 @@ jobs:
...
-          (github.event_name == 'workflow_dispatch' && github.event.inputs.run-tests != 'false')
-          )
+        if: steps.pytest-init-regression.outcome == 'success' || steps.pytest-run-regression.outcome == 'success'
         uses: MishaKav/pytest-coverage-comment@v1.1.51
         with:
           pytest-coverage-path: ./test_coverage.txt
-          junitxml-path: ./test_output.xml
+          junitxml-path: ./pytest-regression/test_output.xml
           create-new-comment: true
+          title: "K8s Regression Test Coverage"
```
ff137, line 806: I don't think the changes to the coverage comment have worked. If you look at the comment (https://github.com/didx-xyz/aries-cloudapi-python/pull/806#issuecomment-2146975836) it no longer contains a coverage report or % test coverage.

We in any case don't need a coverage report for regression tests, and it should instead be based on the CICD / Test results, instead of Deploy to EKS. I guess that can be changed in another PR, so will approve

ff137 (approved):

.github/workflows/continuous-deploy.yml
wdbasson: Should have fixed it now in anyway. Forgot to update the `pytest-coverage-path`.

cl0ete (approved):

PR #808, opened 2024-05-17
:art: Update trust registry models

ff137:
Previously trust registry models were defined in two locations: one place using TypedDicts, the other pydantic BaseModels. Now we consistently use the pydantic models, and routes that say they return model types now actually do (no longer returns dicts instead)


cl0ete (approved):
LGTM

PR #814, opened 2024-05-21
📝 Update wallet credentials docstrings

cl0ete:
Update `wallet/credentails` docstrings

Update route to list `w3c` credentials. It classes with `.../credentials/{credential_id}`

Extend models to have a `credential_id` field


app/routes/wallet/credentials.py
```diff
@@ -136,38 +246,96 @@ async def get_credential_revocation_status(
...
+    ---
+        contexts: Optional[List[str]]
+        types: Optional[List[str]]
+        schema_ids: Optional[List[str]]
+        issuer_id: Optional[str]
+        subject_ids: Optional[List[str]]
+        given_id: Optional[str]
+        proof_types: Optional[List[str]]
+        tag_query: Optional[str]
+        max_results: Optional[int]
```
ff137, line 282: In hierdie  gevalle waar alles optional is, verkies ek om net "Parameters" rename na "Optional Parameters", and die Optional[..] te drop

ff137 (changes requested):

app/models/wallet.py
```diff
@@ -1,27 +1,29 @@
...
 
 class VCRecordList(BaseModel):
-    results: Optional[List[ExtendedVCRecord]] = None
+    results: Optional[List[VCRecord]] = None
 
 
-class ExtendedIndyCredInfo(IndyCredInfo):
+class IndyCredInfo(IndyCredInfoAcaPy):
     credential_id: str = Field(..., alias="referent")
     referent: str = Field(..., alias="credential_id")
```
ff137, line 25: The double aliasing is a bit confusing to me. e.g. if referent already exists in the AcaPy model, I don't see why we redefine it, aliasing the field which is the alias of the referent ... confusing. Can't we just define credential_id? 

I also think it's worthwhile to add a """description""" to the class detailing why we are masking the acapy model. In this case it's just to rename a field to credential_id

app/routes/wallet/credentials.py
```diff
@@ -38,8 +33,9 @@ async def list_credentials(
     Fetch a list of credentials from the wallet
     ---
 
-    The WQL or wallet query language parameter can be used to filter credentials returned from the wallet.
-    The WQL query is a string that can be used to filter records based on the attributes name and value of the record.
+    The WQL (or, wallet query language) parameter can be used to filter credentials returned from the wallet.
+    The `wql` query parameter is a string that can be used to filter records based
+    on the attributes name and value of the record.
```
ff137, line 38: Some duplication here: 
- "The WQL (...) parameter can be used to filter ..."
- "The `wql` query parameter ... can be used to filter ..."

We can probably just have one line, combining the definition of what WQL is, and the fact that `wql` takes a string...

Note: `(or, wallet query language)` can just be `(Wallet Query Langage)`

app/routes/wallet/credentials.py
```diff
@@ -1,34 +1,64 @@
...
+    Fetch a list of credentials from the wallet
+    ---
+
+    The WQL (or, wallet query language) parameter can be used to filter credentials returned from the wallet.
+    The `wql` query parameter is a string that can be used to filter records based
+    on the attributes name and value of the record.
+
+    The following string will look for the credential with the attribute `age` with value `21`:
+
+        {"attr::age::value": "21"}
```
ff137, line 42: I think WQL is important enough to show at least a couple different examples of how it can be used.

We show it can fetch credentials by value. It can also fetch credentials with a specific attribute name? Is there also a way of getting credentials where age >= 21? Let's try add examples for those as well if possible

app/routes/wallet/credentials.py
```diff
@@ -45,12 +75,31 @@ async def list_credentials(
...
 async def get_credential_record(
     credential_id: str,
     auth: AcaPyAuth = Depends(acapy_auth_from_header),
 ) -> IndyCredInfo:
-    """Fetch a specific credential by ID."""
+    """
+    Fetch a specific credential by credential ID
+    ---
+
+    The referent and credential_id are the duplicates of each other.
```
ff137, line 91: Not a necessary detail. What's more important is the `referent` is deprecated. But remember that these descriptions must be relevant to the function of the route.

app/routes/wallet/credentials.py
```diff
@@ -109,14 +189,40 @@ async def get_credential_mime_types(
...
     auth: AcaPyAuth = Depends(acapy_auth_from_header),
 ) -> CredRevokedResult:
-    """Query the revocation status of a specific credential by ID."""
+    """
+    Query the revocation status of a specific credential by ID
+    ---
+
+    The status can be check on a specific time range by providing the `from_` and `to` parameters.
+    Pass the seconds, since Unix epoch, to the `from_` and `to` parameters to query the revocation
+    status for a specific time range.
```
ff137, line 209: "The status can be check" -> The revocation status of a credential can be queried over a specific time range by passing unix timestamps to the `from_` and `to` parameters. Leaving these parameters blank will return the current revocation status.

app/routes/wallet/credentials.py
```diff
@@ -136,38 +242,96 @@ async def get_credential_revocation_status(
...
+    given_id: Optional[str] = Query(None),
+    tag_query: Optional[str] = Query(None),
+    max_results: Optional[int] = Query(None),
     auth: AcaPyAuth = Depends(acapy_auth_from_header),
 ) -> VCRecordList:
-    """Fetch a list of W3C credentials from the wallet."""
+    """
+    Fetch a list of W3C credentials from the wallet
+    ---
+    Credential_id and record_id are duplicates of each other.
```
ff137, line 265: Again, this is a "dangling" detail. It's not tied together with where these fields are found, or why they are duplicated. It can just be omitted.

app/routes/wallet/credentials.py
cl0ete: Removed link from docstrings as I was unable to reproduce any of their examples. Only the one above.

app/routes/wallet/credentials.py
```diff
@@ -136,38 +235,96 @@ async def get_credential_revocation_status(
...
-    body: Optional[W3CCredentialsListRequest] = None,
+    contexts: Optional[List[str]] = Query(None),
+    types: Optional[List[str]] = Query(None),
+    schema_ids: Optional[List[str]] = Query(None),
+    subject_ids: Optional[List[str]] = Query(None),
+    proof_types: Optional[List[str]] = Query(None),
+    issuer_id: Optional[str] = Query(None),
+    given_id: Optional[str] = Query(None),
+    tag_query: Optional[str] = Query(None),
+    max_results: Optional[int] = Query(None),
```
ff137, line 252: Let's trim all these options down to: schema_ids, maybe issuer_id, and max_results

Question - does issuer_id == issuer_did? Let's check in ACA-Py or the responses if that's the case ... because we can name it `issuer_did` for consistency.

Similarly, I guess `max_results` is like `limit`. So let's call it limit for consistency. I know that ACA-Py needs a contribution here for the pagination feature too, but that can come later

app/routes/wallet/credentials.py
```diff
@@ -136,38 +235,96 @@ async def get_credential_revocation_status(
...
+    ---
+        contexts: List[str]
+        types: List[str]
+        schema_ids: List[str]
+        issuer_id: str
+        subject_ids: List[str]
+        given_id: str
+        proof_types: List[str]
+        tag_query: str
+        max_results: int
```
ff137, line 271: After the options are trimmed, we can just add a quick description for each one we keep here.

app/routes/wallet/credentials.py
```diff
@@ -1,34 +1,59 @@
...
 
-@router.get("", response_model=CredInfoList)
+@router.get(
+    "",
+    response_model=CredInfoList,
+    summary="Fetch a list of credentials from the wallet",
+)
 async def list_credentials(
     count: Optional[str] = None,
     start: Optional[str] = None,
```
ff137, line 28: We can rename this limit / offset, and just pass it as the count / start arguments.
I have an open PR with acapy to rename these, but we can rename it on our side so long. And they should be Optional[int]

app/models/wallet.py
```diff
@@ -1,5 +1,37 @@
...
+    results: Optional[List[VCRecord]] = None
+
+
+class IndyCredInfo(IndyCredInfoAcaPy):
+    credential_id: str = Field(
+        ..., alias="referent", description="Rename referent to credential_id"
+    )
+    referent: str = Field(
+        ..., alias="credential_id", description="For backwards compatibility"
+    )
```
ff137, line 33: I think the field descriptions here should be something else.

We're overriding the ACA-Py models, so for our new name (`credential_id`), let's use the ACA-Py description of record_id/referent (the one we're renaming), and then we keep the same description for record_id/referent (which we are preserving for backward compatibility), and just add in brackets (deprecated - renamed to credential_id). Something like that. Also see if Field() accepts a deprecated=True for those

app/routes/wallet/credentials.py
```diff
@@ -1,34 +1,59 @@
...
+        wql: str
+            A WQL query to filter records.
+
+    Returns:
+    ---
+        CredInfoList
+            A list of credential records.
+
+    """
     logger.info("GET request received: List credentials")
```
ff137, line 57: ah, just check that we rebase this to be .debug() log levels - I think all of the received/success logs in this module should just be debug level

ff137 (changes requested):

PR #817, opened 2024-05-24
📝 Update docstrings wallet/dids

cl0ete:
Update `wallet/did` docstrings

Create DID endpoint request model now extends AcaPy model, to flatten `options` field. This brings `key_type` and `did` fields up. 

⚠️ Deprecates the `options` field in the DIDCreate request body, and moves the nested `key_type` and `did` fields up one level.

### DIFF
--- a/app/models/wallet.py
+++ b/app/models/wallet.py
@@ -1,14 +1,11 @@
+import warnings
 from typing import List, Optional
 
+from aries_cloudcontroller import DIDCreateOptions
 from aries_cloudcontroller.models.indy_cred_info import (
     IndyCredInfo as IndyCredInfoAcaPy,
 )
 from aries_cloudcontroller.models.vc_record import VCRecord as VCRecordAcaPy
-from pydantic import BaseModel, Field
-import warnings
-from typing import Optional
-
-from aries_cloudcontroller import DIDCreateOptions
 from pydantic import BaseModel, Field, StrictStr, model_validator
 from typing_extensions import deprecated
 
--- a/app/models/wallet.py
+++ b/app/models/wallet.py
@@ -51,7 +51,7 @@ class CredInfoList(BaseModel):
 
 class DIDCreate(BaseModel):
     method: Optional[StrictStr] = Field(
-        default=None,
+        default="sov",
         description="Method for the requested DID. Supported methods are 'key', 'sov', and any other registered method.",
         examples=["sov", "key", "did:peer:2", "did:peer:4"],
     )
--- a/app/models/wallet.py
+++ b/app/models/wallet.py
@@ -54,7 +54,9 @@ class DIDCreate(DIDCreateAcaPy):
 
     method: Optional[StrictStr] = Field(
         default="sov",
-        description="Method for the requested DID. Supported methods are 'sov', `web`, `did:peer:2` or `did:peer:4`.",
+        description=(
+            "Method for the requested DID. Supported methods are 'sov', 'key', 'web', 'did:peer:2', or 'did:peer:4'."
+        ),
         examples=["sov", "key", "web", "did:peer:2", "did:peer:4"],
     )
     options: Optional[DIDCreateOptions] = Field(
@@ -72,10 +74,9 @@ class DIDCreate(DIDCreateAcaPy):
         description="Key type to use for the DID key_pair. Validated with the chosen DID method's supported key types.",
         examples=["ed25519", "bls12381g2"],
     )
-    did: Optional[str] = Field(
+    did: Optional[StrictStr] = Field(
         default=None,
         description="Specify the final value of DID (including `did:<method>:` prefix) if the method supports it.",
-        strict=True,
     )
 
     @model_validator(mode="before")
--- a/app/models/wallet.py
+++ b/app/models/wallet.py
@@ -47,8 +47,9 @@ class CredInfoList(BaseModel):
 
 class DIDCreate(DIDCreateAcaPy):
     """
-    Extends the AcapyDIDCreate model with smart defaults and simplified interface.
-    Automatically handles the options field structure while maintaining compatibility.
+    Extends the AcapyDIDCreate model with smart defaults and a simplified interface.
+    Handles deprecated `options` field from client requests by populating `key_type` and `did`.
+    Downstream processes should use the appropriate `options` structure based on the model's fields.
     """
 
     method: Optional[StrictStr] = Field(
@@ -59,7 +60,7 @@ class DIDCreate(DIDCreateAcaPy):
     options: Optional[DIDCreateOptions] = Field(
         default=None,
         deprecated=True,
-        description="To define a key type and/or a did depending on chosen DID method.",
+        description="(Deprecated) Define a key type and/or a DID depending on the chosen DID method.",
         examples=[{"key_type": "ed25519", "did": "did:peer:2"}],
     )
     seed: Optional[StrictStr] = Field(
@@ -73,7 +74,7 @@ class DIDCreate(DIDCreateAcaPy):
     )
     did: Optional[str] = Field(
         default=None,
-        description="Specify final value of did (including did:<method>: prefix) if the method supports/requires it.",
+        description="Specify the final value of DID (including `did:<method>:` prefix) if the method supports it.",
         strict=True,
     )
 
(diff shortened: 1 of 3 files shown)


app/routes/wallet/dids.py
```diff
@@ -19,12 +19,30 @@
...
     did_create: Optional[DIDCreate] = None,
     auth: AcaPyAuth = Depends(acapy_auth_from_header),
 ):
-    """Create Local DID."""
+    """
+    Create Local DID
+    ---
+
+    This endpoint allows you to create a new DID in the wallet.
+    The `method` parameter is optional and can be set to 'key' or 'sov'.
```
ff137, line 32: We now also support: `web`, `did:peer:2`, `did:peer:4`

app/routes/wallet/dids.py
```diff
@@ -19,12 +19,30 @@
...
+
+    Request body:
+    ---
+        DIDCreate:
+            method: Optional[str]
+            options: Optional[DIDCreateOptions]
```
ff137, line 38: There's one "gotcha" to keep in mind with the `DIDCreateOptions`

They are optional, but it contains a required field. So, if a client passes `.., options: {}, ..`, then they'll get a validation error, since `key_type` is required and must be one of `["ed25519", "bls12381g2"]`.

Maybe we can streamline this somehow ... instead of us accepting `options: Optional[DIDCreateOptions]`, let's just expose the 2 fields that it contains separately. So we expose:
```
    did: Optional[Annotated[str, Field(strict=True)]] = Field(
        default=None,
        description="Specify final value of the did (including did:<method>: prefix)if the method supports or requires so.",
    )
    key_type: StrictStr = Field(
        description="Key type to use for the DID keypair. Validated with the chosen DID method's supported key types."
    )
```

And we have a default for key_type = ed25519, indicating it's required, and did is optional if they want to set its final value.

I think unwrapping the options field like that will make things a bit clearer for the client

Then of course we have to think about backward compatibility, since replacing options is a breaking change.
So, see if you can keep options - mark it as deprecated - but add `did` and `key_type` as additional args, which configures the DIDCreateOptions we pass to acapy, unless the deprecated `options` field was passed, in which case we use that. Later we sunset and remove the options field.

app/routes/wallet/dids.py
```diff
@@ -19,12 +19,30 @@
 router = APIRouter(prefix="/v1/wallet/dids", tags=["wallet"])
 
 
-@router.post("", response_model=DID)
+@router.post("", response_model=DID, summary="Create Local DID")
 async def create_did(
     did_create: Optional[DIDCreate] = None,
     auth: AcaPyAuth = Depends(acapy_auth_from_header),
 ):
```
ff137, line 26: ) -> DID:
( maybe will be included with rebase)

app/routes/wallet/dids.py
```diff
@@ -19,12 +19,30 @@
...
+    Request body:
+    ---
+        DIDCreate:
+            method: Optional[str]
+            options: Optional[DIDCreateOptions]
+            seed: Optional[str]
+
+    Response:
+    ---
+        Returns the created DID.
```
ff137, line 43: created DID object.

app/routes/wallet/dids.py
```diff
@@ -37,12 +55,19 @@ async def create_did(
...
     """
-    Retrieve list of DIDs.
+    Retrieve list of DIDs
+    ---
+
+    This endpoint allows you to retrieve a list of DIDs in the wallet.
+
+    Response:
+    ---
+        Returns a list of DIDs.
```
ff137, line 70: list of DID objects

app/routes/wallet/dids.py
```diff
@@ -133,13 +210,27 @@ async def get_did_endpoint(
...
-@router.post("/{did}/endpoint", status_code=204)
+@router.post("/{did}/endpoint", status_code=204, summary="Set DID Endpoint")
 async def set_did_endpoint(
     did: str,
     body: SetDidEndpointRequest,
     auth: AcaPyAuth = Depends(acapy_auth_from_header),
 ) -> None:
-    """Update Endpoint in wallet and on ledger if posted to it."""
+    """
+    Update Endpoint in wallet and on ledger if posted to it
```
ff137, line 220: Update endpoint of DID in wallet (and on ledger, if it is a public DID)

app/routes/wallet/dids.py
```diff
@@ -60,12 +85,20 @@ async def list_dids(
...
+@router.get("/public", response_model=DID, summary="Fetch Public DID")
 async def get_public_did(
     auth: AcaPyAuth = Depends(acapy_auth_from_header),
 ) -> DID:
     """
-    Fetch the current public DID.
+    Fetch the current public DID
+    ---
+
+    This endpoint allows you to fetch the current public DID.
```
ff137, line 96: If the description is identical to the heading, then we don't need the description.

We can add some tidbit note: by default, only issuers have public DIDs.

app/routes/wallet/dids.py
```diff
@@ -83,12 +116,27 @@ async def get_public_did(
...
     did: str,
     auth: AcaPyAuth = Depends(acapy_auth_from_header),
 ) -> DID:
-    """Set the current public DID."""
+    """
+    Set the current public DID
+    ---
+
+    This endpoint allows you to set the current public DID.
+    The tenant needs a connection with the endorser to make a did public.
```
ff137, line 129: Let's say "with an endorser", and include a note somewhere (by default, only issuers have a connection with the endorser, and so only issuers can update their public did)

app/routes/wallet/dids.py
```diff
@@ -99,28 +147,57 @@ async def set_public_did(
...
+    bound_logger.info("PATCH request received: Rotate key pair for DID")
     async with client_from_auth(auth) as aries_controller:
-        bound_logger.debug("Rotating keypair")
+        bound_logger.debug("Rotating key pair")
         await handle_acapy_call(
             logger=logger, acapy_call=aries_controller.wallet.rotate_keypair, did=did
         )
 
-    bound_logger.info("Successfully rotated keypair.")
+    bound_logger.info("Successfully rotated key pair.")
```
ff137, line 178: Ensure that rebasing includes all these info logs have been moved to debug level

ff137 (changes requested):

app/models/wallet.py
```diff
@@ -41,3 +44,59 @@ class IndyCredInfo(IndyCredInfoAcaPy):
...
+        Priority: If both are provided, new fields take precedence.
+        """
+        options = values.get("options")
+
+        if options:
+            warnings.warn(
+                "The 'options' field is deprecated. Please use 'key_type' and 'did' fields directly.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
```
ff137, line 90: This will just log a deprecation warning in our stack, which I guess we don't want. We want the user to know it's deprecated, so marking in the model (and our changelogs) should be enough

app/tests/routes/wallet/dids/test_create_did.py
```diff
@@ -1,21 +1,33 @@
...
+        (None, None),
+        (
+            DIDCreate(method="key"),
+            AcapyDIDCreate(method="key", options={"key_type": "ed25519"}),
+        ),
+        (
+            DIDCreate(method="sov"),
+            AcapyDIDCreate(method="sov", options={"key_type": "ed25519"}),
+        ),
+    ],
```
ff137, line 28: The unit tests are cheap so it's good to cover almost every scenario. May as well branch this out to cover more method/key_type combinations

app/routes/wallet/dids.py
```diff
@@ -82,12 +129,28 @@ async def get_public_did(
...
     auth: AcaPyAuth = Depends(acapy_auth_from_header),
 ) -> DID:
-    """Set the current public DID."""
+    """
+    Set the current public DID
+    ---
+
+    This endpoint allows you to set the current public DID.
+    The tenant needs a connection with an endorser to make a did public and by default
+    only issuers will have public DIDs and so only issuers can update their public did.
```
ff137, line 143: Run-on sentence, and mixes capital DIDs with lower case dids. Just refactor this a lil.

app/tests/routes/wallet/dids/test_create_did.py
```diff
@@ -1,21 +1,94 @@
...
+            ),
+        ),
+        (
+            DIDCreate(method="web", key_type="bls12381g2", did="did:web:1234"),
+            DIDCreate(
+                method="web",
+                options={"key_type": "bls12381g2", "did": "did:web:1234"},
+                key_type="bls12381g2",
+                did="did:web:1234",
+            ),
```
ff137, line 87: Just to double check -- the first DIDCreate is the request we get in from the client, and the 2nd one is what we pass to the acapy agent -- through the acapy_wallet.create_did method.  

But, acapy only knows about the options field - not the extra key_type + did fields. They'll just be ignored.

So it'll be more consistent to ensure that the method being passed to acapy just has the method and options

app/models/wallet.py
```diff
@@ -41,3 +43,56 @@ class IndyCredInfo(IndyCredInfoAcaPy):
...
+
+        if values.get("options") is None:
+            values["options"] = {}
+            values["options"]["key_type"] = values.get("key_type") or "ed25519"
+            values["options"]["did"] = values.get("did")
+        else:
+            if not values["options"]["key_type"]:
+                values["options"]["key_type"] = values.get("key_type") or "ed25519"
+            if not values["options"].get("did") and values.get("did"):
+                values["options"]["did"] = values.get("did")
```
ff137, line 96: If we're in the else block, then options is None, so the secondary if checks are redundant. Funny enough, they're not safe! Since the top if calls .get("options") and here the keys are just being accessed, so it could raise a KeyError.

I think we can just drop the secondary `if not` statements.

And `values.get("key_type") or "ed25519"` can just be `values.get("key_type", "ed25519")`

app/routes/wallet/dids.py
```diff
@@ -19,13 +19,38 @@
 router = APIRouter(prefix="/v1/wallet/dids", tags=["wallet"])
 
 
-@router.post("", response_model=DID)
+@router.post("", response_model=DID, summary="Create Local DID")
 async def create_did(
     did_create: Optional[DIDCreate] = None,
```
ff137, line 24: Note: this model is our custom model. And when it's passed to the acapy_wallet service:
```py
        result = await acapy_wallet.create_did(
            did_create=did_create, controller=aries_controller
        )
```
that service is expecting an ACA-Py model. So for consistency, we should just modify our custom model, and ensure we're passing the ACA-Py model to the service. It's just about ensuring we're not passing extra fields which will be ignored -- safer to ensure we're passing acapy model to acapy.

app/routes/wallet/dids.py
```diff
@@ -19,13 +19,38 @@
...
+            seed: [str]
+            key_type: [str]
+            did: [str]
+
+    Response:
+    ---
+        Returns the created DID object.
+
+    """
+    logger.debug("POST request received: Create DID {}", did_create)
```
ff137, line 53: I think if we want to log the request body, we should use a bound_logger, so the following logger.debug lines share the same context.
So it'll be bound_logger = logger.bind(body = did_create), and update to use bound_logger in this method. Something like that

ff137 (changes requested):

app/routes/wallet/dids.py
cl0ete: I'm gonna remove this. Was using it for some debugging

app/models/wallet.py
cl0ete: Hey I updated the logic here let me know what you think

ff137 (approved):

PR #824, opened 2024-05-27
📝 💥Swagger tenant admin

cl0ete:
Update docstrings

Deprecated  `GET .../v1/tenants/{wallet_id}/access-token`

Added `POST .../v1/tenants/{wallet_id}/access-token`

Breaking changes:
 - Delete tenant endpoint from `200` to `204` status code

### DIFF
--- a/app/routes/admin/tenants.py
+++ b/app/routes/admin/tenants.py
@@ -240,8 +240,9 @@ async def delete_tenant_by_id(
     Delete Tenant by ID
     ---
 
-    Use this endpoint to delete a Tenant by its Wallet ID. This will remove the Tenant's Wallet and any associated
-    credentials, connections, etc. And delete them from the trust registry if they are an issuer or verifier.
+    Use this endpoint to delete a Tenant by its Wallet ID. This action will remove the Tenant's Wallet,
+    along with any associated credentials, connections, and other data. If the tenant is an issuer or verifier,
+    they will also be removed from the trust registry.
 
     Request parameters:
     ---
@@ -304,7 +305,7 @@ async def get_wallet_auth_token(
     Request parameters:
     ---
         wallet_id: str
-            The Wallet ID of the Tenant for which to rotate the access token.
+            The Wallet ID of the tenant for which the access token will be rotated.
 
     Response body:
     ---
@@ -355,7 +356,7 @@ async def get_wallet_auth_token(
     Request parameters:
     ---
         wallet_id: str
-            The Wallet ID of the Tenant for which to rotate the access token..
+            The Wallet ID of the tenant for which the access token will be rotated.
 
     Response body:
     ---
@@ -398,16 +399,16 @@ async def update_tenant(
     Update Tenant by Wallet ID
     ---
 
-    Use this endpoint to update a Tenant's details based on their Wallet ID.
+    Update a Tenant's details based on their Wallet ID.
 
-    Holders cannot have their roles updated. Attempting to assign issuer or
-    verifier roles to a holder will result in a 409 conflict error.
-
+    Holders cannot have their roles updated. Attempting to assign issuer or verifier
+    roles to a holder will result in a 409 conflict error.
+    
     Only issuers or verifiers can be updated to hold both roles.
 
     This endpoint does not support revoking roles from issuers or verifiers.
 
-    For issuers and verifiers, updates to `image_url` will be reflected on the trust registry.
+    Updates to `image_url` for issuers and verifiers will be reflected on the trust registry.
 
 
     Request body:
--- a/app/routes/admin/tenants.py
+++ b/app/routes/admin/tenants.py
@@ -95,7 +95,7 @@ async def create_tenant(
             extra_settings: Optional[Dict[str, Union[bool, str]]]
                 Optional per-tenant settings to configure wallet behaviour for advanced users.
 
-    Response body:(Paginated)
+    Response body:
     ---
         CreateTenantResponse
             wallet_id: str
@@ -105,7 +105,6 @@ async def create_tenant(
             image_url: Optional[str]
             updated_at: str
             access_token: str
-            group_id: Optional[str]
     """
     bound_logger = logger.bind(body=body)
     bound_logger.debug("POST request received: Starting tenant creation")
--- a/app/routes/admin/tenants.py
+++ b/app/routes/admin/tenants.py
@@ -67,8 +67,15 @@ async def create_tenant(
     Create a new tenant
     ---
 
-    Use this endpoint to create a new tenant. The tenant will be created with a wallet and
-    optionally onboarded with roles.
+    Use this endpoint to create a new tenant.
+    This will create a new wallet (the tenant) and onboard it with the requested roles.
+
+    The `wallet_name` is an optional field that allows you to assign a name to the wallet,
+    useful with `get_tenants` to fetch wallets by wallet name.
+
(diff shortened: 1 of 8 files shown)


app/routes/admin/tenants.py
```diff
@@ -58,12 +58,58 @@
...
+            wallet_name: Optional[str]
+                An optional wallet name. Useful with `get_tenants` to fetch wallets by wallet name.
+            roles: Optional[List[str]]
+                A list of roles to assign to the tenant.
+            image_url: Optional[str]
+                An optional image URL for the tenant.
+            group_id: Optional[str]
+                An optional group identifier.
+            extra_settings: Optional[Dict[str, Union[bool, str]]]
+                Optional per-tenant settings to configure wallet behaviour for advanced users.
```
ff137, line 99: Ask GPT to proof read / improve some of this Maximising conciseness. We're duplicating a lot of info from the request body descriptions. We should just add noteworthy things here (no roles = regular holder. image_url is added to the trust registry if they have a role). And I don't think we should refer to `tenant-admin` here, since the reader will presumably be an admin.

app/routes/admin/tenants.py
```diff
@@ -187,13 +233,28 @@ async def create_tenant(
     return response
 
 
-@router.delete("/{wallet_id}")
+@router.delete("/{wallet_id}", summary="Delete a Tenant by wallet ID", status_code=204)
```
ff137, line 236: Consistent capitalisation for key words. `.. by Wallet ID`

app/routes/admin/tenants.py
```diff
@@ -187,13 +233,28 @@ async def create_tenant(
...
     admin_auth: AcaPyAuthVerified = Depends(acapy_auth_tenant_admin),
-):
-    """Delete tenant by id."""
+) -> None:
+    """
+    Delete tenant by ID
+    ---
+
+    Use this endpoint to delete a tenant by its wallet ID. This will remove the tenant's wallet and any associated
+    credentials, connections, etc.
```
ff137, line 247: And delete them from the trust registry if they are an issuer or verifier.

app/routes/admin/tenants.py
```diff
@@ -256,14 +337,46 @@ async def get_wallet_auth_token(
     return response
 
 
-@router.put("/{wallet_id}", response_model=Tenant)
+@router.put("/{wallet_id}", response_model=Tenant, summary="Update Tenant by wallet ID")
```
ff137, line 340: `by Wallet ID` (elsewhere too)

app/routes/admin/tenants.py
```diff
@@ -226,12 +287,32 @@ async def delete_tenant_by_id(
         bound_logger.debug("Successfully deleted tenant.")
 
 
-@router.get("/{wallet_id}/access-token", response_model=TenantAuth)
+@router.get(
```
ff137, line 290: Note to self: this should really be a PATCH endpoint, so it's more clear that this modifies existing tokens.

Maybe we can mark this one as deprecated, and copy it identically with the one modified to being a `patch` endpoint

One of those things that should just be a done...

app/routes/admin/tenants.py
```diff
@@ -226,12 +287,32 @@ async def delete_tenant_by_id(
...
+    response_model=TenantAuth,
+    summary="Update auth token by wallet ID",
+)
 async def get_wallet_auth_token(
     wallet_id: str,
     group_id: Optional[str] = group_id_query,
     admin_auth: AcaPyAuthVerified = Depends(acapy_auth_tenant_admin),
 ) -> TenantAuth:
+    """
+    Get wallet access token by id
```
ff137, line 301: Rotate access token (instead of Get)

app/routes/admin/tenants.py
```diff
@@ -256,14 +337,46 @@ async def get_wallet_auth_token(
...
     admin_auth: AcaPyAuthVerified = Depends(acapy_auth_tenant_admin),
 ) -> Tenant:
-    """Update tenant by id."""
+    """
+    Update tenant by id
+    ---
+
+    Use this endpoint to update a tenant by its wallet ID.
+
+    A Holder can't have their roles updated. Only Issuers and Verifiers can have their roles updated.
```
ff137, line 353: To be clear: a holder cannot be _promoted_ to an issuer or verifier. Trying to promote a holder will return a 409. Only issuers or verifiers can be promoted to both roles. This endpoint can't be used to revoke roles from an issuer or verifier.
^ let's try paraphrase that ... with GPT's help :-)

Can also note that image_url for issuers/verifiers will be updated on trust registry

app/routes/admin/tenants.py
```diff
@@ -284,13 +397,34 @@ async def update_tenant(
...
     wallet_id: str,
     group_id: Optional[str] = group_id_query,
     admin_auth: AcaPyAuthVerified = Depends(acapy_auth_tenant_admin),
 ) -> Tenant:
-    """Get tenant by id."""
+    """
+    Get tenant by id
+    ---
+
+    Use this endpoint to fetch a tenant by its wallet ID.
```
ff137, line 410: to fetch tenant info

app/routes/admin/tenants.py
```diff
@@ -307,7 +441,7 @@ async def get_tenant(
     return response
 
 
-@router.get("", response_model=List[Tenant])
+@router.get("", response_model=List[Tenant], summary="Get all Tenants")
```
ff137, line 444: Maybe something like: summary="Fetch All Tenants (Paginated)"
and then need query param info added for limit/offset/order_by/descending

app/routes/admin/tenants.py
```diff
@@ -58,12 +58,58 @@
...
+
+    Use this endpoint to create a new tenant.
+    This will create a new wallet (the tenant) and onboard it with the requested roles.
+
+    The `wallet_name` is an optional field that allows you to assign a name to the wallet,
+    useful with `get_tenants` to fetch wallets by wallet name.
+
+    `image_url` is an optional field that allows tenant-admin to assign an image to the wallet.
+
+    `group_id` is an optional field that allows the tenant-admin to assign the wallet to a group.
```
ff137, line 78: Note: the admin can't control group_id. The group_id **query param** has `include_in_schema=False`, but it does still appear in request bodies... hmm, so we perhaps we should mark it as hidden in the request body too. AFAIK from Robbie's work we inject group_id to request bodies, and will override whatever the client enters there. So, it shouldn't be in the model or docstrings imo

ff137 (changes requested):

app/routes/admin/tenants.py
```diff
@@ -256,14 +387,52 @@ async def get_wallet_auth_token(
...
+    Update a Tenant's details based on their Wallet ID.
+
+    Holders cannot have their roles updated. Attempting to assign issuer or verifier
+    roles to a holder will result in a 409 conflict error.
+
+    Only issuers or verifiers can be updated to hold both roles.
+
+    This endpoint does not support revoking roles from issuers or verifiers.
+
+    Updates to `image_url` for issuers and verifiers will be reflected on the trust registry.
```
ff137, line 410: NB: Does updating wallet_label get reflected on the trust registry? I haven't checked, but noting down as important question

ff137: Ok, it does. ✅

ff137 (approved):
🎉

PR #842, opened 2024-06-19
Create a service for definitions endpoints

cl0ete:
Moved bulk of `app/routes/definitions.py` to `app/services/definitions.py`


app/services/definitions.py
```diff
@@ -0,0 +1,543 @@
...
+    credential_schema_from_acapy,
+)
+from app.util.retry_method import coroutine_with_retry_until_value
+from shared import ACAPY_ENDORSER_ALIAS, REGISTRY_CREATION_TIMEOUT
+from shared.log_config import get_logger
+
+logger = get_logger(__name__)
+
+
+@dataclass
```
ff137, line 44: Hmm ... if we follow this pattern, it's much simpler to just have:

```py
class SchemaPublisher:
    def __init__(self, logger: Logger, controller: AcaPyClient):
        self._logger = logger
        self._controller = controller
```

No custom data class needed, and more straight forward reference to logger or controller

app/services/definitions.py
```diff
@@ -0,0 +1,543 @@
...
+                error.status_code,
+                error.detail,
+            )
+            if error.status_code == 405:
+                self.deps.logger.info(
+                    "Status code 405 indicates schema is already registered, so we can continue"
+                )
+            else:
+                raise error
+
```
ff137, line 164: If we have multiple classes for our service, I think it's better for each class to be in its own module.

Then again, having a class SchemaRegistrar, for encapsulating one method, feels like we can just have naked method `register_schema`, without the class pattern, like most of our code! I actually prefer having classes instead of naked methods, but it only makes sense if there's a lot of complexity and class helps simplify things

For instance, here, `registrar = SchemaRegistrar(deps)` is just giving us access to this one method, which we could instead just call by passing controller and logger. Logger doesn't even have to be passed. It's only when we want to maintain the bound context. Since we are going from route to service, it's probably ok for service to just define it's own module's logger at the top, we can .bind whatever we want here

app/services/definitions.py
```diff
@@ -0,0 +1,543 @@
...
+        trust_registry_schemas = [await get_trust_registry_schema_by_id(schema_id)]
+
+    schema_ids = [schema.id for schema in trust_registry_schemas]
+
+    bound_logger.debug("Getting schemas associated with fetched ids")
+    schemas = await schema_futures(schema_ids, aries_controller)
+
+    if schema_issuer_did:
+        schemas = [
+            schema for schema in schemas if schema.id.split(":")[0] == schema_issuer_did
```
ff137, line 236: technically `[0]` is unsafe and may raise a IndexError. Good practice to wrap anything that can raise an exception in a try-block

app/services/definitions.py
```diff
@@ -0,0 +1,543 @@
...
+        raise CloudApiException(
+            "An unexpected error occurred: could not publish schema."
+        )
+
+    result = credential_schema_from_acapy(result.sent.var_schema)
+    bound_logger.info("Successfully published and registered schema.")
+    return result
+
+
+async def get_schemas_tenant(
```
ff137, line 204: `get_schemas_as_tenant`

app/services/definitions.py
```diff
@@ -0,0 +1,543 @@
...
+        ]
+    if schema_name:
+        schemas = [schema for schema in schemas if schema.name == schema_name]
+    if schema_version:
+        schemas = [schema for schema in schemas if schema.version == schema_version]
+
+    return schemas
+
+
+async def get_schemas_governance(
```
ff137, line 246: `get_schemas_as_governance`

app/services/definitions.py
```diff
@@ -0,0 +1,543 @@
...
+    aries_controller: AcaPyClient,
+    schema_id: Optional[str],
+    schema_issuer_did: Optional[str],
+    schema_name: Optional[str],
+    schema_version: Optional[str],
+) -> List[CredentialSchema]:
+    """
+    Governance agents gets all schemas created by itself
+    """
```
ff137, line 255: There should be some assertion at the start of this method which asserts that the controller is a governance controller. Otherwise this method can be called using any controller

app/services/definitions.py
```diff
@@ -0,0 +1,543 @@
...
+    # Initiate retrieving all schemas
+    schema_ids = response.schema_ids or []
+
+    bound_logger.debug("Getting schemas associated with fetched ids")
+    schemas = await schema_futures(schema_ids, aries_controller)
+
+    return schemas
+
+
+async def schema_futures(
```
ff137, line 284: I think futures refers to the awaitable before it's awaited. So this method can have a different name. Like get_schemas_by_id

app/services/definitions.py
```diff
@@ -0,0 +1,543 @@
...
+    schema_ids: List[str], aries_controller: AcaPyClient
```
ff137, line 285: Aah let's maintain pattern everywhere of having the controller as the first argument

app/services/definitions.py
```diff
@@ -0,0 +1,543 @@
...
+                )
+
+            else:
+                self.deps.logger.error(log_message)
+                client_error_message = (
+                    "Something went wrong while asserting if request is from a valid issuer. "
+                    "Please try again."
+                )
+            raise CloudApiException(client_error_message, e.status_code) from e
+        return public_did
```
ff137, line 348: asserting public did should be a util method of sorts, reusable anywhere, not specific to this class

app/services/definitions.py
```diff
@@ -0,0 +1,543 @@
...
+    schemas = [
+        credential_schema_from_acapy(schema.var_schema)
+        for schema in schema_results
+        if schema.var_schema
+    ]
+
+    return schemas
+
+
+class CredDefPublisher:
```
ff137, line 321: See this class makes sense -- I just think maybe `check_endorser_connection` can be moved to a util method as well, because there may be more contexts that also need to assert a controller has a connection with the endorser -- so it can be re-used, not just in creating cred defs. Same with wait for transaction ack, I can imagine that being reused elsewhere too. But wait_for_rev_reg, perfect, makes sense for that to be part of a CredDefPublisher. I would just move it to its own module, and perhaps rename to CredentialDefinitionPublisher. Long name, but the keystrokes are free, and I think verbosity (within reason) can help with maintainability somewhat

ff137 (changes requested):
The different classes needs to be consolidated. Some can just be raw methods. Some make sense to move to their own modules. Some other methods can be moved to utils. 

We don't use the class pattern a lot so far, but I think it is better, and generally good practice to have related methods together in a relevant class, as opposed to one giant service module with all the methods we need. Logger / controller can just be used directly

app/services/definitions.py
cl0ete: @ff137  Chatting to Claude and reading the python docs it looks like `.split()` will always return a list with at least one element *if* a delimiter is passed to `.splt()`. 
- ` "".split(":")` returns `['']`
- `":123".split(":")` returns `['', '123']

Given our models and the validator on the schema id for the TR this should be safe.
My Claude chat:
Screenshot from 2024-06-26 09-11-02

ff137 (changes requested):
Looking good - just a variety of little change requests

app/services/definitions/schema_publisher.py
```diff
@@ -0,0 +1,106 @@
...
+        if set(_schema.var_schema.attr_names) != set(schema.attribute_names):
+            error_message = (
+                "Error creating schema: Schema already exists with different attribute names. "
+                f"Given: `{str(set(schema.attribute_names))}`. "
+                f"Found: `{str(set(_schema.var_schema.attr_names))}`."
+            )
+            raise CloudApiException(
+                error_message,
+                409,
+            )
```
ff137, line 99: can be 1 line, just drop the last , and auto-format

app/services/definitions/schema_publisher.py
```diff
@@ -0,0 +1,106 @@
...
+from app.models.definitions import CreateSchema, CredentialSchema
+from app.util.definitions import credential_schema_from_acapy
+
+
+class SchemaPublisher:
+    def __init__(
+        self,
+        controller: AcaPyClient,
+        logger: Logger,
+    ):
```
ff137, line 17: think this can also fit in 1 line. nitpick 🎨 
adding comma at end forces black to have it multiline, without it it will add the comma if it can't fit in 1 line

ff137 (changes requested):
only minor changes required

app/services/definitions/definitions.py
```diff
@@ -0,0 +1,325 @@
...
+        await register_schema(schema_id=result.sent.schema_id)
+    else:
+        bound_logger.error("No SchemaSendResult in `publish_schema` response.")
+        raise CloudApiException(
+            "An unexpected error occurred: could not publish schema."
+        )
+
+    result = credential_schema_from_acapy(result.sent.var_schema)
+    bound_logger.info("Successfully published and registered schema.")
+    return result
```
ff137, line 77: I think this method needs some refactoring.
- First, the method name can just be `create_schema`, instead of `create_schema_service`.
- We want to make sure that only governance can create_schemas. Even if we know that check is happening in the route logic, there should be an assertion here as well, to check that aries_controller is indeed the governance controller, and it can't accidentally be called elsewhere as a tenant.
- Also, I think the whole try publish, if already exists call handle existing, and if result, publish to trust registry, I think that can all be part of the SchemaPublisher. So the SchemaPublisher has a .create_schema method, which does all of this, and this definitions method just forwards the request to the schema publisher, and does the `credential_schema_from_acapy` conversion. And maybe the SchemaSendRequest model creation too. Something like that. That way, `handle_existing_schema` can be an internal method (`_handle_existing_schema`), not called from outside.

Lastly, I'm wondering why the method takes both a `SchemaSendRequest` and a `CreateSchema` in its argument ... `CreateSchema` is the model we receive as request body, and SchemaSendRequest is what we send to ACA-Py. It feels like only one of those is needed here. Lots of ways to skin a cat. Maybe just move the SchemaSendRequest model creation here. Or modify handle_existing_schema to not need the CreateSchema body ... see what works best!

app/services/definitions/credential_definition_publisher.py
```diff
@@ -0,0 +1,95 @@
...
+from app.util.check_endorser_connection import check_endorser_connection
+from shared import REGISTRY_CREATION_TIMEOUT
+
+
+class CredentialDefinitionPublisher:
+    def __init__(self, controller: AcaPyClient, logger: Logger):
+        self._logger = logger
+        self._controller = controller
+
+    async def assert_public_did(self):
```
ff137, line 18: It doesn't look like this method is called ... remember, there is now util.assert_public_did, which is what's called in services.definitions.create_cred_def:
```
    publisher = CredentialDefinitionPublisher(
        controller=aries_controller, logger=bound_logger
    )

    public_did = await assert_public_did(aries_controller)
```

Is that meant to call this publisher.assert_public_did instead? Or is one of the methods not needed anymore?

app/services/definitions/definitions.py
```diff
@@ -0,0 +1,325 @@
...
+            aries_controller=aries_controller, transaction_id=result.txn.transaction_id
+        )
+
+    if support_revocation:
+        await publisher.wait_for_revocation_registry(credential_definition_id)
+
+    return credential_definition_id
+
+
+async def get_cred_defs(
```
ff137, line 262: I think full names are better. `get_credential_definitions`

app/services/definitions/definitions.py
```diff
@@ -0,0 +1,325 @@
...
+    schemas = [
+        credential_schema_from_acapy(schema.var_schema)
+        for schema in schema_results
+        if schema.var_schema
+    ]
+
+    return schemas
+
+
+async def create_cred_def(
```
ff137, line 213: `create_credential_definition`

ff137 (changes requested):

app/services/definitions/credential_definition_publisher.py
cl0ete: I forgot to remove :facepalm:

ff137 (approved):
I included some minor :art:, mainly splitting `services.definitions.definitions.py` into `schemas.py` and `credential_definitions.py`, and then importing as `schemas_service` and `cred_def_service` in the routes module

PR #850, opened 2024-06-27
:sparkles: Pagination support for fetching wallets, connections, cred ex and proof records

### LINKED ISSUE #621: Paginate long responses
Some routes can potentially return very large responses, such as:
- list tenants
- list connection records
- list exchange / presentation records

It would make sense to review such cases and add features that enable pagination for these endpoints. This will help with handling large datasets or high traffic.

Suggested library: https://github.com/uriyyo/fastapi-pagination

ff137:
Closes #621

Adds `limit` and `offset` query parameters to endpoints for:
- fetching wallets
- fetching connection records
- fetching v1 or v2 credential exchange records
- fetching v1 or v2 presentation exchange records

Todo:
- [x] add more e2e tests, asserting pagination works as expected

Later todo:
- add custom ordering support, when added in askar

### DIFF
--- a/.github/workflows/codacy.yml
+++ b/.github/workflows/codacy.yml
@@ -40,7 +40,7 @@ jobs:
 
       # Execute Codacy Analysis CLI and generate a SARIF output with the security issues identified during the analysis
       - name: Run Codacy Analysis CLI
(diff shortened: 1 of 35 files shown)


cl0ete (approved):
GGWP

PR #902, opened 2024-07-24
:wastebasket: Remove v1 protocol version and default to v2

ff137:
- 🔥 removes the v1 issuer and verifier services
- 🎨 removes "protocol version" from request bodies, as we now only support v2
- ✅ update tests to accommodate to these changes


cl0ete (approved):
LGTM - We should just consider how this would affect an existing wallet with a credential issued with the v1 protocol... :thinking:

PR #906, opened 2024-07-25
⬆️ 👷  Dependabot Groups

rblaine95:
* Add all python modules to the python dependabot group
* Hopefully will reduce dependabot spam


rblaine95:
It looks like Dependabot Groups will do multiple dependencies in a single PR for multiple groups.

What we want is 1 PR per dependency across the modules.
So, if `fastapi` updates, instead of 4 PRs to update `fastapi`, we just want 1 PR to update all the modules for that 1 update.

Issue #899, opened 2024-07-24
⚡ Improve backfilling of SSE events with more scalable solution


ff137:
When there are many wallets, the SSE backfilling logic during the startup of the webhooks service is simply not efficient or scalable.

Currently, we fetch all wallet ids, and then for each one we query if it has any events within a recent timespan. This is to pre-fill SSE queues in the case where webhooks service has had to restart, so that we don't drop recent events for SSE subscribers.

A more scalable approach would be to simply write events to a separate redis key, with a time-to-live/expiry equal to the duration that we want to backfill. So that upon startup, we can fetch all the keys that have not expired, and then we immediately knows these are the wallet ids with recent events in our window of interest.

ff137:
Closed by #1095

PR #961, opened 2024-08-07
:adhesive_bandage: Remove `_backfill_events` method from webhooks service startup

ff137:
Backfilling SSE queues with recent events is currently highly inefficient when many wallets exist. Fortunately this is a non-critical feature, and so can just be removed for now. A solution will be implemented as detailed in https://github.com/didx-xyz/aries-cloudapi-python/issues/899


ff137:
Codacy is currently down with an outage for one of their services. Merging as minor changes don't need codacy review

PR #964, opened 2024-08-08
Increase wait for issuer did transaction to 30s

wdbasson:


ff137 (approved):
Note: the timeout will now be 15s. Since it's 30 attempts * 0.5s retry delay.

cl0ete (approved):

PR #976, opened 2024-08-12
Configure global concurrency

wdbasson:
Solve issue where `deploy` job from a concurrent workflow run takes the concurrency lock before the `test` job of a run was able to lock it. Which meant tests weren't being executed against the actual assets relevant to their commit.

### DIFF
--- a/.github/workflows/continuous-deploy.yml
+++ b/.github/workflows/continuous-deploy.yml
@@ -33,7 +33,12 @@ on:
       - synchronize
       - ready_for_review
 
+concurrency:
+  group: cicd
+  cancel-in-progress: false
+
 jobs:
+
   build:
     if: ${{ !github.event.pull_request.draft }}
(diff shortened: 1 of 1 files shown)


rblaine95 (approved):

ff137 (approved):

ff137 (commented):
Not sure but the Deploy to EKS job here seems it may have cancelled my EKS / pytest run here: https://github.com/didx-xyz/aries-cloudapi-python/actions/runs/10351777014/job/28651869843?pr=975

wdbasson:
> Not sure but the Deploy to EKS job here seems it may have cancelled my EKS / pytest run here: https://github.com/didx-xyz/aries-cloudapi-python/actions/runs/10351777014/job/28651869843?pr=975

Ah, yes it would have because that job was still using the old concurrency group so the new workflow run will have interfered.

PR #979, opened 2024-08-13
🧑‍💻 Remove patch proxy CICD logic

wdbasson:
Patching the agents with RDS Proxy config is now handled via helm `--state-values-set rdsProxyEnabled=true`.

Removes the need for the patch proxy step which saves some time and reduces overhead.

Also executing "regression" tests before "clean" tests since they take longer.

### DIFF
--- a/.github/workflows/continuous-deploy.yml
+++ b/.github/workflows/continuous-deploy.yml
@@ -387,24 +387,6 @@ jobs:
           authkey: ${{ secrets.TAILSCALE_AUTHKEY }}
           version: ${{ env.TAILSCALE_VERSION }}
 
-      # - name: Helmfile Diff
-      #   if: github.event.inputs.run-reset-deployments == 'true'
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous-deploy.yml
```diff
@@ -484,7 +466,7 @@ jobs:
         env:
           IMAGE_TAG: ${{ needs.build.outputs.image_version }}
 
-      - name: Helmfile Apply (Exclude Agents) # Always exclude agents and apply (zero diff if resetting deployments)
+      - name: Helmfile Apply (RDS Proxy) # Always exclude agents and apply (zero diff if resetting deployments)
```
rblaine95, line 469:
```suggestion
      - name: Helmfile Apply (RDS Proxy)
```

rblaine95 (approved):
Small nitpick, but otherwise LGTM 👍

PR #983, opened 2024-08-15
:sparkles: Add issuer/verifier `image_url` to trustregistry

cl0ete:
Add `image_url` to actor models
Handel new `image_url` in update tenant call and update call on TR

Fix/update tests
Update pre-commit

Added `alembic` a lightweight database migration tool for usage with SQLAlchemy
Did the following for setup:
- ran:  `alembic init migrations`
-  updated `sqlalchemy.url` in `alembic.ini` 
-  import `Base` from `trustregistry/db.py` to `trustregistry/migrations/env.py` 
     set `target_metadata = Base.metadata` in env.py (alembic can see models now)
-  ran: `alembic revision --autogenerate -m "Actor image_url"` to create migration.
  Can see migrations generated in `trustregistry/migrations/versions`

Can now run:
- `alembic upgrade head` to run migration
or
-  `alembic downgrade -1` to undo last migration


ff137 (commented):
What is `alembic`? 

Is the script just for a once-off migration, and can then be removed?

trustregistry/pyproject.toml
```diff
@@ -28,6 +29,7 @@ pylint = "~3.3.0"
 pytest = "~8.3.2"
 pytest-cov = "~5.0.0"
 pytest-mock = "~3.14.0"
+alembic = "~1.13.2"
```
ff137, line 32: can be removed here

trustregistry/tests/test_main.py
```diff
@@ -19,28 +19,221 @@ def test_create_app():
...
+            "head_rev",
+            False,
+        ),  # alembic_version exists, revisions don't match
+        (
+            True,
+            True,
+            "same_rev",
+            "same_rev",
+            True,
+        ),  # alembic_version exists, revisions match
```
ff137, line 165: if the comment is on the same line, then formatting expands it to this multiline view. Maybe comments can be added to line before, so it's more compact

trustregistry/migrations/README
```diff
@@ -0,0 +1 @@
+Generic single-database configuration.
```
ff137, line 1: This can either be removed or just edited with a one line summary of what this directory is for

trustregistry/main.py
```diff
@@ -19,10 +24,59 @@
 ROOT_PATH = os.getenv("ROOT_PATH", "")
 
 
+def check_migrations(
+    engine: Engine, alembic_cfg: Config  # pylint: disable=redefined-outer-name
```
ff137, line 28: This `redefined-outer-name` is because we name the param `engine`, and also import `from trustregistry.database import engine`. Better would be to rename one of those two.

ff137 (approved):

PR #984, opened 2024-08-15
Simple graceful shutdown

wdbasson:


app/main.py
```diff
@@ -83,7 +84,7 @@ async def lifespan(_: FastAPI):
     # Shutdown logic occurs after yield
     logger.info("Calling WebsocketManager shutdown")
     await WebsocketManager.disconnect_all()
-
+    time.sleep(30)
```
ff137, line 87: `await asyncio.sleep(30)` instead

30 maybe excessive tho?

wdbasson: > `await asyncio.sleep(30)` instead
> 
> 30 maybe excessive tho?

Currently handling this with pre-stop lifecycle hooks. For synchronous calls we might need longer sleep periods to accommodate long response times, _e.g._, creating credential definitions that wait for revocation registries.

PR #989, opened 2024-08-19
✨ Add ordering functionality to fetching of askar records

### LINKED ISSUE #908: ✨ Support ordering in paginated queries
Continuing from #621, it is necessary for us to support custom ordering in our paginated querying of storage records, such that the end user has more control over results appearing on page 1, but also to ensure that DB shuffling does not result in records being moved across pages, resulting in duplicate records across pages, or records being missed as offset is incremented.

This requires support in Askar: https://github.com/hyperledger/aries-askar/issues/290
and in ACA-Py: https://github.com/hyperledger/aries-cloudagent-python/issues/3001

ff137:
testing ordering functionality introduced on ACA-Py test branch: https://github.com/hyperledger/aries-cloudagent-python/pull/3173

adds `order_by` and `descending` options when fetching tenants, connection, or exchange records.

`order_by` has a default value of the record id column. This is the only valid value, and effectively works like ordering by creation time. So the main functionality is in the `descending: boolean` option, and being able to paginate with guarantee of record positions.

Closes #908


cl0ete (approved):
LGTM. Yes just the `limit` > `0`

PR #990, opened 2024-08-19
:arrow_up: :sparkles:  Upgrade acapy 1.0.0

cl0ete:
Enrich revoked responses

   - Add get pending revocations endpoint (per registry)
   - Publish revocation endpoint now returns cred_rev_ids per rev_reg_id published
   - Revoke true will also return cred_rev_ids per rev_reg_id published
   - added and fixed tests

Upgrade to Aca-Py v1.0.0

### DIFF
--- a/app/models/issuer.py
+++ b/app/models/issuer.py
@@ -101,16 +101,19 @@ class ClearPendingRevocationsResult(BaseModel):
 
 
 class RevokedResponse(BaseModel):
-    result: Dict[str, List[int]] = Field(
+    cred_rev_ids_published: Dict[str, List[int]] = Field(
         default_factory=dict,
-        description="A map of revocation registry IDs to lists of credential revocation IDs (as integers) that have been revoked. Can be empty.",
+        description=(
+            "A map of revocation registry IDs to lists of credential revocation IDs"
+            "(as integers) that have been revoked. Can be empty."
+        ),
     )
 
     @root_validator(pre=True)
     def extract_revoked_info(cls, values):
         txn = values.get("txn", {})
         messages_attach = txn.get("messages_attach", [])
-        result = {}
+        cred_rev_ids_published = {}
 
         for attach in messages_attach:
             data = attach.get("data", {}).get("json", {})
@@ -119,9 +122,9 @@ class RevokedResponse(BaseModel):
             revoked = operation.get("value", {}).get("revoked", [])
 
             if revoc_reg_def_id and revoked:
-                result[revoc_reg_def_id] = revoked
+                cred_rev_ids_published[revoc_reg_def_id] = revoked
 
-        values["result"] = result
+        values["cred_rev_ids_published"] = cred_rev_ids_published
         return values
 
 
--- a/app/models/issuer.py
+++ b/app/models/issuer.py
@@ -110,6 +110,7 @@ class RevokedResponse(BaseModel):
     )
 
     @model_validator(mode="before")
+    @classmethod
     def extract_revoked_info(cls, values):
         txn = values.get("txn", {})
         messages_attach = txn.get("messages_attach", [])
--- a/app/models/issuer.py
+++ b/app/models/issuer.py
@@ -101,4 +101,25 @@ class ClearPendingRevocationsResult(BaseModel):
 
 
 class RevokedResponse(BaseModel):
-    revoked_cred_rev_ids: list[Optional[int]] = []
+    result: Dict[str, List[int]] = Field(
+        default_factory=dict,
+        description="A map of revocation registry IDs to lists of credential revocation IDs (as integers) that have been revoked. Can be empty.",
+    )
+
+    @root_validator(pre=True)
+    def extract_revoked_info(cls, values):
+        txn = values.get("txn", {})
+        messages_attach = txn.get("messages_attach", [])
+        result = {}
+
+        for attach in messages_attach:
+            data = attach.get("data", {}).get("json", {})
+            operation = data.get("operation", {})
+            revoc_reg_def_id = operation.get("revocRegDefId")
+            revoked = operation.get("value", {}).get("revoked", [])
+
+            if revoc_reg_def_id and revoked:
+                result[revoc_reg_def_id] = revoked
+
+        values["result"] = result
+        return values
--- a/app/models/issuer.py
+++ b/app/models/issuer.py
@@ -2,7 +2,7 @@ from enum import Enum
 from typing import Dict, List, Optional
 
 from aries_cloudcontroller import LDProofVCDetail
-from pydantic import BaseModel, Field, ValidationInfo, field_validator, root_validator
+from pydantic import BaseModel, Field, ValidationInfo, field_validator, model_validator
 
 from shared.exceptions import CloudApiValueError
 from shared.models.protocol import IssueCredentialProtocolVersion
@@ -109,7 +109,7 @@ class RevokedResponse(BaseModel):
         ),
     )
 
-    @root_validator(pre=True)
+    @model_validator(mode="before")
     def extract_revoked_info(cls, values):
         txn = values.get("txn", {})
         messages_attach = txn.get("messages_attach", [])
--- a/app/models/issuer.py
+++ b/app/models/issuer.py
@@ -1,7 +1,7 @@
 from enum import Enum
 from typing import Any, Dict, List, Optional
 
-from aries_cloudcontroller import LDProofVCDetail
+from aries_cloudcontroller import LDProofVCDetail, TxnOrPublishRevocationsResult, TransactionRecord
 from pydantic import BaseModel, Field, ValidationInfo, field_validator, model_validator
 
 from shared.exceptions import CloudApiValueError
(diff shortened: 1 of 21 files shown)


dockerfiles/agents/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ghcr.io/hyperledger/aries-cloudagent-python:py3.12-1.0.0rc4
+FROM ghcr.io/hyperledger/aries-cloudagent-bbs:py3.12-1.0.0
```
rblaine95, line 1: Do we use BBS for anything?
It sounds like we support BBS for JSONLD credentials 🤔 
But JSONLD might not be functional yet?

It looks like Hyperledger has been working on reworking/reimplementing BBS to be ARM compatible.
* https://github.com/hyperledger/aries-bbssignatures-rs/issues/8
* https://github.com/mattrglobal/ffi-bbs-signatures/pull/74
* https://github.com/mattrglobal/ffi-bbs-signatures/pull/77

If we can drop BBS then we can enable multi-arch builds on our Agents.
At least until the rework/reimplementation of BBS is done.

Alternatively, we keep ARM disabled for the agents until BBS is ARM compatible, at which point we then enable ARM builds.

@ff137, what do you think?

From a personally self-interested standpoint, I'd really like to get ARM compatibility sooner rather than later.
The only things currently in this repo that aren't ARM compatible are:
* Aca-Py Agents
* Ledger Nodes/Browser

ff137: > But JSONLD might not be functional yet?
We have signing and verifying of JSON-LD documents, which does need to be reviewed again to ensure everything is working properly.

But, we support BBS for sending `ld_proof` credentials. So, the issuer can have a did:key with BBS method, and send ld_proof credentials specifying their did. We have tests that assert that works. Other than that, we don't really use BBS.

So, we could disable support for BBS in the meantime. Mainly just need to skip the relevant tests. And then re-enable those when BBS is ARM-compatible.

app/models/issuer.py
```diff
@@ -98,3 +102,39 @@ class ClearPendingRevocationsResult(BaseModel):
...
         ),
     )
+
+
+class RevokedResponse(BaseModel):
+    cred_rev_ids_published: Dict[str, List[int]] = Field(
+        default_factory=dict,
+        description=(
+            "A map of revocation registry IDs to lists of credential revocation IDs "
+            "(as integers) that have been revoked. Can be empty."
```
ff137, line 112: "Can be empty if ..." We should say what an empty list means. In this context, an empty list of a cred_rev_ids means no creds were revoked for that rev_reg_id?

app/models/issuer.py
```diff
@@ -98,3 +102,39 @@ class ClearPendingRevocationsResult(BaseModel):
...
+        ),
+    )
+
+    @model_validator(mode="before")
+    @classmethod
+    def extract_revoked_info(
+        cls, values: TxnOrPublishRevocationsResult
+    ) -> Dict[str, Any]:
+        if isinstance(values, dict) and "txn" in values:
+            txn_list: List[TransactionRecord] = values.get("txn")
```
ff137, line 122: minor note: this isn't actually a List[TransactionRecord]. It's a List[Dict[str,Any]], which has same shape as List[TransactionRecord]. Point being that the pydantic model doesn't have a .get method. So the type hint makes it seem like `txn.get` will error. So, for clarity we can just set the type hint to list of dict, and comment that it's list transactionrecord

app/routes/issuer.py
```diff
@@ -523,20 +525,24 @@ async def revoke_credential(
...
         bound_logger.debug("Revoking credential")
-        await revocation_registry.revoke_credential(
+        result = await revocation_registry.revoke_credential(
             controller=aries_controller,
             credential_exchange_id=body.credential_exchange_id,
             auto_publish_to_ledger=body.auto_publish_on_ledger,
         )
 
-    bound_logger.debug("Successfully revoked credential.")
+    bound_logger.info("Successfully revoked credential.")
```
ff137, line 544: should be debug log

app/routes/issuer.py
```diff
@@ -681,7 +693,8 @@ async def publish_revocations(
                     504,
                 ) from e
 
-    bound_logger.debug("Successfully published revocations.")
+    bound_logger.info("Successfully published revocations.")
```
ff137, line 696: should be debug log

app/routes/issuer.py
```diff
@@ -739,3 +752,39 @@ async def clear_pending_revocations(
...
+    bound_logger = logger.bind(body={"revocation_registry_id": revocation_registry_id})
+    bound_logger.info("GET request received: Get pending revocations")
+
+    async with client_from_auth(auth) as aries_controller:
+        bound_logger.debug("Getting pending revocations")
+        result = await revocation_registry.get_pending_revocations(
+            controller=aries_controller, rev_reg_id=revocation_registry_id
+        )
+
+    bound_logger.info("Successfully fetched pending revocations.")
```
ff137, line 789: GET and Success logs are now debug level

app/services/revocation_registry.py
```diff
@@ -144,12 +145,27 @@ async def revoke_credential(
...
+
+        if (
+            revoke_result
+            and revoke_result["txn"]
+            and revoke_result["txn"]["messages_attach"][0]
+        ):
+            bound_logger.info("Successfully revoked credential.")
+            return RevokedResponse.model_validate({"txn": [revoke_result["txn"]]})
+
+    bound_logger.info("Successfully revoked credential.")
```
ff137, line 162: debug

app/services/revocation_registry.py
```diff
@@ -181,19 +197,19 @@ async def publish_pending_revocations(
...
+    if not result.txn or not result.txn[0].transaction_id:
         bound_logger.warning(
             "Published pending revocations but received no endorser transaction id. Got result: {}",
             result,
         )
         return
 
-    endorse_transaction_id = result.txn.transaction_id
-    bound_logger.debug(
+    endorse_transaction_id = result.txn[0].transaction_id
```
ff137, line 207: hmm ... looks like we need to now endorse a potential list of transaction ids? Here we're just taking the first one

app/services/revocation_registry.py
```diff
@@ -181,19 +197,19 @@ async def publish_pending_revocations(
...
+    bound_logger.info(
```
ff137, line 208: debug

app/services/revocation_registry.py
```diff
@@ -487,3 +503,38 @@ async def wait_for_active_registry(
...
+            acapy_call=controller.revocation.get_registry,
+            rev_reg_id=rev_reg_id,
+        )
+    except CloudApiException as e:
+        raise CloudApiException(
+            f"Failed to get pending revocations: {e.detail}", e.status_code
+        ) from e
+
+    pending_revocations = result.result.pending_pub
+    bound_logger.info("Successfully retrieved pending revocations.")
```
ff137, line 539: debug logs

app/services/revocation_registry.py
```diff
@@ -487,3 +503,38 @@ async def wait_for_active_registry(
...
+
+    Args:
+        controller (AcaPyClient): aca-py client
+        rev_reg_id (str): The revocation registry ID.
+
+    Raises:
+        Exception: When the pending revocations could not be retrieved.
+
+    Returns:
+        pending_revocations (List[str]): The pending revocations.
```
ff137, line 522: Mismatch - method says returns List[int], description says List[str]

Also make it more clear what "The pending revocations" is - it's cred_rev_ids?

Looks like `result.result.pending_pub` returns List[str]. Should they be transformed to ints?

app/services/revocation_registry.py
```diff
@@ -487,3 +503,38 @@ async def wait_for_active_registry(
...
+            logger=bound_logger,
+            acapy_call=controller.revocation.get_registry,
+            rev_reg_id=rev_reg_id,
+        )
+    except CloudApiException as e:
+        raise CloudApiException(
+            f"Failed to get pending revocations: {e.detail}", e.status_code
+        ) from e
+
+    pending_revocations = result.result.pending_pub
```
ff137, line 538: Note: result.result can be None. So calling .pending_pub could be unsafe.

`pending_revocations = result.result.pending_pub if result.result else []`

app/services/revocation_registry.py
```diff
@@ -144,12 +145,27 @@ async def revoke_credential(
...
+                "Revocation was published but no result was returned. "
+                "Has this credential been revoked before?"
+            )
+
+        if (
+            revoke_result
+            and revoke_result["txn"]
+            and revoke_result["txn"]["messages_attach"][0]
+        ):
+            bound_logger.info("Successfully revoked credential.")
```
ff137, line 159: debug

app/services/revocation_registry.py
```diff
@@ -487,3 +503,38 @@ async def wait_for_active_registry(
...
+        rev_reg_id (str): The revocation registry ID.
+
+    Raises:
+        Exception: When the pending revocations could not be retrieved.
+
+    Returns:
+        pending_revocations (List[str]): The pending revocations.
+    """
+    bound_logger = logger.bind(body={"rev_reg_id": rev_reg_id})
+    bound_logger.info("Fetching pending revocations for a revocation registry")
```
ff137, line 525: debug

ff137 (commented):
Overall great stuff! Just a couple important changes (avoid potential KeyError; and should we endorse a list of transaction ids now?), and some nitpicks: log levels and type hints

PR #997, opened 2024-08-27
Expose query params for `get_credentials_by_proof_id`

### LINKED ISSUE #991: Match proof with credentials endpoint only returns 10 credentials
`/tenant/v1/verifier/proofs/{{proofId}}/credentials` 

Only returns 10 credentials and is not currently configurable.

cl0ete:
add query params:
- count
- start
- referent  <-- This is the `presentation_referents`. This can be used to specify which of the `presentation_referents` to get credentials for

closes #991


app/routes/verifier.py
```diff
@@ -484,19 +484,29 @@ async def delete_proof(
...
     Parameters:
     ---
         proof_id: str
             The relevant proof exchange ID for the prover
+        referent: Optional str
+            The presentation_referent of the proof to match
+        count: Optional str
+            The number of credentials to fetch
+        start: Optional str
+            The index to start fetching credentials from
```
ff137, line 509: Let's rename `count` and `start` to `limit` and `offset`, on our side. So it's the same as the other endpoints with pagination

ff137 (changes requested):
Let's add a unit test where limit/offset is requested, and assert that the same value is called in the acapy method (for count/start)

Also, maybe let's add an e2e test. Issue 2 credentials. Call the method with count = 1,2, and 3. Assert 1, 2 and 2 are returned. Then call with limit=1 and offset 0,1,2. Assert that 1,1,0 is returned. Something like that.

Maybe also assert behaviour with regards to `referent` being called in the query params.

If you like, e2e tests can be implemented in another PR, but preferably unit tests in this one.

app/routes/verifier.py
```diff
@@ -484,19 +484,29 @@ async def delete_proof(
 )
 async def get_credentials_by_proof_id(
     proof_id: str,
+    referent: Optional[str] = None,
+    limit: Optional[str] = None,
+    offset: Optional[str] = None,
```
ff137, line 489: Not sure if limit and offset are meant to be strings :-)

Also, we can define them as query params with validation: min value = 0 for both. max value = 10'000 for limit. something like that.

Then in unit tests, assert validation error is raised when passing bad limit/offset values

cl0ete: Laat my weet wat jy dink van die toets wat ek ge-add het

app/routes/verifier.py
```diff
@@ -484,19 +484,29 @@ async def delete_proof(
...
     Parameters:
     ---
         proof_id: str
             The relevant proof exchange ID for the prover
+        referent: Optional str
+            The presentation_referent of the proof to match
+        limit: Optional str
+            The number of credentials to fetch
+        offset: Optional str
+            The index to start fetching credentials from
```
ff137, line 509: limit and offset = ints, for the route docs

app/routes/verifier.py
```diff
@@ -512,7 +522,11 @@ async def get_credentials_by_proof_id(
         async with client_from_auth(auth) as aries_controller:
             bound_logger.debug("Fetching credentials for request")
             result = await verifier.get_credentials_by_proof_id(
-                controller=aries_controller, proof_id=proof_id
+                controller=aries_controller,
+                proof_id=proof_id,
+                referent=referent,
+                count=str(limit),
+                start=str(offset),
```
ff137, line 529: Ek sien dat acapy dit as strings verwag ... I'll make a quick PR for them

ff137: https://github.com/hyperledger/aries-cloudagent-python/pull/3208

app/routes/verifier.py
```diff
@@ -484,19 +484,29 @@ async def delete_proof(
 )
 async def get_credentials_by_proof_id(
     proof_id: str,
+    referent: Optional[str] = None,
+    limit: Optional[int] = Query(default=100, ge=0, le=10000),
+    offset: Optional[int] = Query(default=0, ge=0),
```
ff137, line 489: Soz, one last nitpick! We can deduplicate by using the `limit_query_parameter` and `offset_query_parameter` in `app/util/pagination.py`. 

(They don't have the ge/le validation yet - I'll add it there)

cl0ete: cool will do. Also adding the e2e tests

ff137 (approved):

PR #1000, opened 2024-08-29
🦺 🩹  Aca-py V1 update cleanup

cl0ete:
Address some issues found in v1 update

### DIFF
--- a/app/routes/issuer.py
+++ b/app/routes/issuer.py
@@ -671,28 +671,27 @@ async def publish_revocations(
             return RevokedResponse()
 
         endorser_transaction_ids = [txn.transaction_id for txn in result.txn]
-        if endorser_transaction_ids:
-            for endorser_transaction_id in endorser_transaction_ids:
-                bound_logger.debug(
-                    "Wait for publish complete on transaction id: {}",
-                    endorser_transaction_id,
+        for endorser_transaction_id in endorser_transaction_ids:
+            bound_logger.debug(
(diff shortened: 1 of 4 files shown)


app/routes/issuer.py
```diff
@@ -670,30 +670,31 @@ async def publish_revocations(
...
-                await coroutine_with_retry_until_value(
-                    coroutine_func=aries_controller.endorse_transaction.get_transaction,
-                    args=(endorser_transaction_id,),
-                    field_name="state",
-                    expected_value="transaction_acked",
-                    logger=bound_logger,
-                    max_attempts=30,
-                    retry_delay=1,
+        endorser_transaction_ids = [txn.transaction_id for txn in result.txn]
+        if endorser_transaction_ids:
```
ff137, line 674: the if can be dropped, since the for will only iterate if there are elements

app/services/revocation_registry.py
```diff
@@ -204,8 +204,8 @@ async def publish_pending_revocations(
         )
         return
 
-    endorse_transaction_id = result.txn[0].transaction_id
-    bound_logger.info(
+    endorse_transaction_id = [txn.transaction_id for txn in result.txn]
```
ff137, line 207: let's just make it plural: `endorse_transaction_ids`

ff137 (approved):
Nice! Thank you -- just 2 nitpicks

PR #1008, opened 2024-09-06
✨Waypoint

cl0ete:
First draft of Waypoint

TODO:
 - [x] Add health check
    - Might need to improve health checks  
 - [x] Unit tests


waypoint/main.py
```diff
@@ -0,0 +1,63 @@
...
+
+    return application
+
+
+logger.info("Waypoint Service startup")
+
+app = create_app()
+
+
+# TODO - Improve health checks
```
ff137, line 55: let's include this :-)
We can maybe just add a `check_connection` method to NatsEventProcessor. Something which pings the JetStreamContext, if possible. Can be left for later if complicated, but maybe worth including now

health_check from webhooks for inspiration:
```py

@app.get("/health")
@inject
async def health_check(
    acapy_events_processor: AcaPyEventsProcessor = Depends(
        Provide[Container.acapy_events_processor]
    ),
    sse_manager: SseManager = Depends(Provide[Container.sse_manager]),
    billing_manager: BillingManager = Depends(Provide[Container.billing_manager]),
):
    if (
        acapy_events_processor.are_tasks_running()
        and sse_manager.are_tasks_running()
        and billing_manager.are_tasks_running()
    ):
        return {"status": "healthy"}
    else:
        raise HTTPException(
            status_code=503, detail="One or more background tasks are not running."
        )

```

waypoint/routes/sse.py
```diff
@@ -0,0 +1,133 @@
+import asyncio
```
ff137, line 1: Minor naming consistency note: the folder `routes` should be `routers`. Just the convention in other services. Also for tests/routes

ff137:
Just marking as draft to save on EKS runs. Will post review soon

ff137 (approved):
Good stuff! Mostly minor comments -- main thing is see if we can include a health check or not. May as well add something now which just asserts NATS is still connected

waypoint/services/nats_service.py
```diff
@@ -0,0 +1,136 @@
...
+        connect_kwargs = {
+            "servers": [NATS_SERVER],
+        }
+        if NATS_CREDS_FILE:
+            connect_kwargs["user_credentials"] = NATS_CREDS_FILE
+
+            nats_client: NATS = await nats.connect(**connect_kwargs)
+        else:
+            logger.warning("No NATS credentials file found, assuming local development")
+            nats_client: NATS = await nats.connect(**connect_kwargs)
```
ff137, line 35:
```py
        if NATS_CREDS_FILE:
            connect_kwargs["user_credentials"] = NATS_CREDS_FILE
        else:
            logger.warning("No NATS credentials file found, assuming local development")
        
        nats_client: NATS = await nats.connect(**connect_kwargs)
```

waypoint/main.py
```diff
@@ -0,0 +1,89 @@
...
+            )
+    except asyncio.TimeoutError:
+        raise HTTPException(
+            status_code=503,
+            detail={"status": "not ready", "error": "JetStream health check timed out"},
+        )
+    except Exception as e:
+        raise HTTPException(
+            status_code=500, detail={"status": "error", "error": str(e)}
+        )
```
ff137, line 89: raising HTTPException in the try block will cause it to be caught as generic Exception

So, perhaps this should be:
```py
    try:
        jetstream_status = await asyncio.wait_for(
            nats_processor.check_jetstream(), timeout=5.0
        )
    except asyncio.TimeoutError:
        raise HTTPException(
            status_code=503,
            detail={"status": "not ready", "error": "JetStream health check timed out"},
        )
    except Exception as e:
        raise HTTPException(
            status_code=500, detail={"status": "error", "error": str(e)}
        )

    if jetstream_status["is_working"]:
        return {"status": "ready", "jetstream": jetstream_status}
    else:
        raise HTTPException(
            status_code=503,
            detail={"status": "not ready", "jetstream": "JetStream not ready"},
        )
```

waypoint/services/nats_service.py
```diff
@@ -0,0 +1,147 @@
...
+        try:
+            account_info = await self.js_context.account_info()
+            is_working = account_info.streams > 0
+            logger.trace(f"JetStream check completed. Is working: {is_working}")
+            return {
+                "is_working": is_working,
+                "streams_count": account_info.streams,
+                "consumers_count": account_info.consumers,
+            }
+        except Exception:
```
ff137, line 146: It's probably worth logging what these exceptions might be.
Use `logger.exception("Caught exception while checking jetstream status")` to automatically print stack trace, without needing to add `as e` and printing `str(e)`

waypoint/services/nats_service.py
```diff
@@ -0,0 +1,147 @@
...
+            return subscription
+
+        except BadSubscriptionError as e:
+            logger.error(f"BadSubscriptionError subscribing to NATS: {e}")
+            raise
+        except Error as e:
+            logger.error(f"Error subscribing to NATS: {e}")
+            raise
+        except Exception as e:
+            logger.error(f"Unknown error subscribing to NATS: {e}")
```
ff137, line 87: Similar here, we want to probably see a stack trace for unknown exceptions. Just replace with:
```py
except Exception:
            logger.exception("Unknown error subscribing to NATS.")
raise
```

docker-compose.yaml
```diff
@@ -257,6 +257,29 @@ services:
...
+      - host
+      - nats
+    depends_on:
+      - governance-endorser
+    # depends_on:
+    #   - nats-1
+    #   - nats-2
+    #   - nats-3
+    env_file:
+      - environments/waypoint/waypoint.env
```
ff137, line 282: Lastly we'd want to configure the healthcheck here.

Similar to how it's done for webhooks/endorser:
```py
    healthcheck:
      test: ["CMD-SHELL", "curl -f http://localhost:3010/health || exit 1"]
      interval: 30s
      timeout: 10s
      retries: 3
      start_period: 40s
```

docker-compose.yaml
```diff
@@ -257,6 +257,35 @@ services:
...
+    healthcheck:
+      test: ["CMD-SHELL", "curl -f http://localhost:3010/health/ready || exit 1"]
```
ff137, line 284: The port 🙉 waypoint is 3011

waypoint/services/nats_service.py
```diff
@@ -0,0 +1,148 @@
...
+        try:
+            account_info = await self.js_context.account_info()
+            is_working = account_info.streams > 0
+            logger.trace(f"JetStream check completed. Is working: {is_working}")
+            return {
+                "is_working": is_working,
+                "streams_count": account_info.streams,
+                "consumers_count": account_info.consumers,
+            }
+        except Exception:
```
ff137, line 146: Just add `  # pylint: disable=W0718` to the `except Exception` line. To suppress pylint warnings

Please find all `except Exception` blocks in this PR and add pylint disable

waypoint/services/nats_service.py
```diff
@@ -0,0 +1,148 @@
...
+            f"Processing events for group {group_id} and wallet {wallet_id} on topic {topic}"
+        )
+
+        subscription = await self._subscribe(group_id=group_id, wallet_id=wallet_id)
+
+        async def event_generator():
+            end_time = time.time() + duration
+            while not stop_event.is_set():
+                remaining_time = remaining_time = end_time - time.time()
+                logger.trace(f"remaining_time: {remaining_time}")
```
ff137, line 109: Important note: don't use f-string interpolation in logs. Because, trace logs are not evaluated when log level is set to info, but if you use f-string interpolation, then that string is evaluated every time this line gets passed over. if you use the {} interpolation, then it's lazily evaluated, and there's no cost to trace lines when log level is not at trace.

In general, you can use f-string for error logs, since then it doesn't really matter. Those lines are basically always logged in any case. But for trace/debug/info, etc, be sure not to use f-interpolation, since there is unnecessary cost to using it.

So ... please search for `f"` to find all the f-string interpolation used in this PR, and replace with the {} pattern:
```py
logger.trace("remaining_time: {}", remaining_time)
```

waypoint/pyproject.toml
```diff
@@ -0,0 +1,50 @@
+[tool.poetry]
```
ff137, line 1: just call `poetry update` in this folder, and then `poetry lock` to make sure everything here is up to date

waypoint/main.py
```diff
@@ -0,0 +1,90 @@
...
+app = create_app()
+
+
+@app.get("/health/live")
+async def health_live():
+    return {"status": "live"}
+
+
+@app.get("/health/ready")
+@inject
```
ff137, line 64: I'm noticing there's two health endpoints, when we probably only need one? In the waypoint logs I can see both get called every 10 seconds, which I don't understand because docker-compose healthcheck for waypoint just has: `test: ["CMD-SHELL", "curl -f http://localhost:3011/health/ready || exit 1"]`

Not sure - please ask Robbie or Willem what's best practice for health checks here

ff137 (changes requested):

waypoint/tests/test_main.py
```diff
@@ -0,0 +1,59 @@
...
+@pytest.mark.anyio
+async def test_health_live():
+    response = await health_live()
+    assert response == {"status": "live"}
+
+
+@pytest.mark.anyio
+async def test_health_ready():
+    response = await health_ready()
+    assert response == {"status": "ready"}
```
ff137, line 59: Btw, please double check the endorser/webhooks tests on the dependency injection mocking, for testing the health endpoint:
```sh
FAILED waypoint/tests/test_main.py::test_health_ready - fastapi.exceptions.HTTPException: 500: {'status': 'error', 'error': "'Depends' object has no attribute 'check_jetstream'"}
```

waypoint/main.py
rblaine95: In this case:

Liveness - if this passes, it means Uvicorn/FastAPI is correctly configured and everything is good to go.
If this fails, the Kubernetes Liveness Probe will trigger a restart.
If this happens too frequently too quickly, the container enters a CrashLoop.

Readiness - if this passes, it means the connection to NATS is functioning correctly and all is well.
If connection to NATS is lost or misconfigured, container will go into not-ready state.

ff137:
P.S. Looks like the EKS Pytest run needs the waypoint dependencies:
```sh
...
waypoint/services/nats_service.py:7: in <module>
    import nats
E   ModuleNotFoundError: No module named 'nats'
```
I can't see where the install should be configured

ff137 (approved):
🎉 awesome, well done!

PR #1019, opened 2024-09-09
:construction: Refactor endorser

ff137:


endorser/services/endorsement_processor.py
```diff
@@ -79,150 +67,41 @@ def are_tasks_running(self) -> bool:
...
+                    await self._process_endorsement_event(message_data)
+                    await message.ack()
+            except TimeoutError:
+                logger.trace("Timeout fetching messages continuing...")
+                await asyncio.sleep(1)
+            except Exception as e:  # pylint: disable=W0703
+                logger.error("Error processing endorsement event: {}", e)
+                await self._handle_unprocessable_endorse_event(
+                    message_subject, message_data, e
+                )
```
ff137, line 104: Technically not safe... because message_subject is potentially unbound.

e.g. endorser just starts up, `message` has not been defined yet, `subscription.fetch` raises some exception (other than TimeoutError), and then line 102 will raise a ValueError, var not defined. It's good to look out for these possibilities in try-except blocks, where new vars are defined in the try, and then used in the exception handling.

endorser/services/endorsement_processor.py
```diff
@@ -274,16 +153,35 @@ def _handle_unprocessable_endorse_event(
...
+
+    async def _subscribe(self) -> JetStreamContext.PullSubscription:
+        """
+        Subscribes to the NATS subject for endorsement events.
+        """
+        logger.info("Subscribing to NATS subject: cloudapi.aries.events.endorser.>")
+        try:
+            subscribe_kwargs = {
+                "durable": ENDORSER_DURABLE_CONSUMER,
+                "subject": f"{NATS_SUBJECT}.endorser.>",
```
ff137, line 175: subject is logged (hardcoded) and then defined as f"{NATS_SUBJECT}.endorser.>"

We can also use `f"{NATS_SUBJECT}.endorser*"`, since > will capture extra groups, which we don't expect

endorser/services/endorsement_processor.py
```diff
@@ -274,16 +153,35 @@ def _handle_unprocessable_endorse_event(
...
+                "stream": NATS_STREAM,
+            }
+            subscription = await self.jetstream.pull_subscribe(**subscribe_kwargs)
+        except (BadSubscriptionError, Error) as e:
+            logger.error("Error subscribing to NATS subject: {}", e)
+            raise e
+        except Exception as e:
+            logger.error("Unknown error subscribing to NATS subject: {}", e)
+            raise e
```
ff137, line 184: Better to use exception log.
```
except Exception:
    logger.exception("Unknown error subscribing to NATS subject")
    raise
```
And catching base Exception should just include pylint warning ignore

app/tests/e2e/verifier/test_many_revocations.py
```diff
@@ -0,0 +1,217 @@
+import asyncio
```
ff137, line 1: Just a note to self to skip this test, so we can still optionally use it for local testing

ff137:
Included in #1095

PR #1020, opened 2024-09-13
🧑‍💻 Reduce Redis and Postgres resource presets

rblaine95:
* Micro -> Nano
* Also reduce Istio Sidecar CPU Requests from `100m` to `10m`


cl0ete (approved):
LGTM

PR #1048, opened 2024-09-19
:arrow_up: Update dependencies

ff137:
Adds waypoint to update_lock_files script.
Bumps fastapi + pydantic versions in waypont
Updates all lock files


scripts/update_lock_files.sh
```diff
@@ -19,7 +19,7 @@ else
 fi
 
 # Array of submodules
-submodules=("app" "endorser" "trustregistry" "webhooks")
```
rblaine95, line 22: I think it'd be better to move away from `scripts/` and standardize on using [Mise Tasks](https://mise.jdx.dev/tasks/).

ff137: Agreed :+1:

.mise/tasks/poetry/update/waypoint
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+# mise description="Poetry Update waypoint lockfile"
```
rblaine95, line 2: Sorry, I missed a titlecase here 🤦 
```suggestion
# mise description="Poetry Update Waypoint lockfile"
```

PR #1050, opened 2024-09-19
:construction_worker: Add Style Check for Tiltfiles

rblaine95:
* Use Mise to install Python and dependencies to run style checks
* Check Tiltfile formatting

---

This also speeds up the Style Check action from ±20s to ±15s.
Probably because of Mise Caching and `uv` is faster than `pip`


ff137 (approved):
🎉

PR #1067, opened 2024-09-23
:test_tube: point e2e tests to use waypoint

cl0ete:
Making sure all `check_webhook_state` functions use the `filter_map` field. 
Waypoint needs a `field` and `field_id`


ff137 (commented):
There's something wrong with the redis nats/benthos pipelines... I think the redis-sync job gets bloated and slows everything down (the e2e issuer tests take 52 minutes to complete, and the ones that error work fine if they are the first to be run). So, we'll leave this open until we're completely ready to remove redis and switch to nats_events

ff137:
Included in #1095

PR #1073, opened 2024-09-26
➕🗃️Alembic bootstrap the db

cl0ete:
Add alembic

Create initial migration that will bootstrap the tables (actors and schemas)

Modify `main.py ` alembic runs the migrations on startup, start with initial migration to create tables


cl0ete:
Added in PR983

PR #1075, opened 2024-09-27
:heavy_plus_sign: :memo: Disable Redoc, replace Swagger with Scalar

rblaine95:
* Disable Redoc (it was never officially used anyways)
* Replace Swagger with Scalar

---

✨  It's pretty ✨ 
image
image


ff137 (approved):
:tada:

PR #1099, opened 2024-10-09
:fire: :coffin: Remove Redis and Webhooks Web from Tilt

rblaine95:
* Remove Redis and Webhooks Web from Tilt
* Make Endorser depend on NATS


cl0ete (approved):

rblaine95:
EKS tests are failing, most likely due to the Alembic PR and DB migrations.

This doesn't affect EKS at all, so I think this is safe to merge @ff137 @cl0ete

PR #1100, opened 2024-10-09
:construction_worker: Use Github Org `CHARTS_PAT` secret

rblaine95:
* Instead of repo managed secret for cloning Charts, use Org secret


ff137 (approved):
Seems good! 
Can maybe be tested by pushing this branch with another name:
`dependabot/github_actions/SonarSource/sonarcloud-github-action-3.1.0`
to match the cloudapi dependabot PR, and see if it passes

PR #1137, opened 2024-10-24
:zap: Waypoint start time

cl0ete:
Add `deliver_policy` to waypoint `by_start_time`
Removed env var `LOOK_BACK`,  add `look_back` query param to sse endpoints. 

Default `look_back` is 300 seconds

### DIFF
--- a/waypoint/services/nats_service.py
+++ b/waypoint/services/nats_service.py
@@ -15,6 +15,8 @@ from shared.models.webhook_events import CloudApiWebhookEventGeneric
 
 logger = get_logger(__name__)
 
+look_back = int(os.getenv("LOOK_BACK", 30))
+
 
 class NatsEventsProcessor:
     """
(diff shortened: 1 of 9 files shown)


waypoint/services/nats_service.py
```diff
@@ -40,7 +45,22 @@ async def _subscribe(
                     "subject": f"{NATS_SUBJECT}.*.{wallet_id}",
```
ff137, line 45: PS, some cleanup we can do for this above:
- remove empty line after `if group_id`
- logs should be more accurate: we don't know if it's a tenant-admin call, we can just say: "Subscribe with group_id: {}", or "Subscribe without group_id". Plus those logs are probably unnecessary.
- define `subscribe_kwargs = {"stream": NATS_STREAM]}` once

Altogether:
```py
        try:
            logger.trace("Subscribing to JetStream for wallet_id: {}, group_id: {}", wallet_id, group_id}
            group_id = group_id or "*"
            subscribe_kwargs = {
                "stream": NATS_STREAM,
                "subject": f"{NATS_SUBJECT}.{group_id}.{wallet_id}",
            }
```

ff137 (approved):

PR #1141, opened 2024-10-28
:sparkles: Implement retry logic in RichAsyncClient

ff137:
Since 503 errors can sometimes occur in our EKS tests, this inspired that our RichAsyncClient implementation should be configurable for retries

Current new defaults: up to 3 retries on 503 errors.


cl0ete (approved):
GGWP

PR #1143, opened 2024-10-30
⚡️ add state monitoring stream

henrymsiska:
Updated nats-setup job for new stream.

Waypoint consumes from new NATS state stream


rblaine95 (approved):

cl0ete (approved):

waypoint/services/nats_service.py
```diff
@@ -73,6 +73,7 @@ async def process_events(
         group_id: str,
         wallet_id: str,
         topic: str,
+        state: str,
```
ff137, line 76: We can add `state` to the logger.debug body

waypoint/services/nats_service.py
```diff
@@ -25,7 +25,7 @@ def __init__(self, jetstream: JetStreamContext):
         self.js_context: JetStreamContext = jetstream
 
     async def _subscribe(
-        self, group_id: str, wallet_id: str, look_back: int
+        self, group_id: str, wallet_id: str, topic: str, state: str, look_back: int
```
ff137, line 28: I think it's worth forcing the naming of args when there's so many in a method.

Meaning, we can add `self, *, group_id: str, ...` and then when _subscribe is called, the args have to be named by keywords. This makes it safer when passing 4 strings in a row, so there's more certainty when passing, and developer doesn't have to double check that args line up correctly.

waypoint/services/nats_service.py
```diff
@@ -73,6 +73,7 @@ async def process_events(
         group_id: str,
```
ff137, line 73: Same here - add `*` to force passing args by keywords

ff137 (approved):

PR #1160, opened 2024-11-04
Test 2 second timeout

wdbasson:
Debugging endorser stuff...

### DIFF
--- a/endorser/services/endorsement_processor.py
+++ b/endorser/services/endorsement_processor.py
@@ -86,7 +86,7 @@ class EndorsementProcessor:
         subscription = await self._subscribe()
         while True:
             try:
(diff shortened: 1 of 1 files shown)


endorser/services/endorsement_processor.py
```diff
@@ -86,7 +86,7 @@ async def _process_endorsement_requests(self) -> NoReturn:
         subscription = await self._subscribe()
         while True:
             try:
-                messages = await subscription.fetch(batch=1, timeout=60, heartbeat=1)
+                messages = await subscription.fetch(batch=1, timeout=2, heartbeat=1)
```
ff137, line 89: btw, endorser logs show: `heartbeat value too large`3

PR #1166, opened 2024-11-05
👷  Validate PR Title Length

rblaine95:
* PR Title length should not exceed 50 characters (excluding emojis)
* Emojis are validated against gitmoji.dev

---

image

### DIFF
--- a/.github/workflows/validate-pr.yml
+++ b/.github/workflows/validate-pr.yml
@@ -108,13 +108,24 @@ jobs:
               comment.body.includes('Your PR title is too long!')
             );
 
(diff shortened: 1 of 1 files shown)


cl0ete (approved):

ff137 (approved):
Nicely done - good workflow to have for our repos in general 👍

rblaine95:
@ff137 for the future, I'll look at making it so that the bot only ever has a single comment that it updates/deletes instead of comment per validation.

I'll also look at putting the action into its own repo where we can then tag it and use it across all our repos and toggle validations.

ff137:
@rblaine95 Epic, great stuff!

PR #1169, opened 2024-11-08
✅ Test limit and offset params on get credentials

cl0ete:
add fixture and test

### DIFF
--- a/app/tests/e2e/test_wallet_credentials.py
+++ b/app/tests/e2e/test_wallet_credentials.py
@@ -69,7 +69,7 @@ async def test_get_and_delete_credential_record(
     TestMode.regression_run in TestMode.fixture_params,
     reason="Skipping due to regression run",
 )
-@pytest.mark.parametrize("issue_alice_many_creds", [3], indirect=True)
+@pytest.mark.parametrize("issue_alice_many_creds", [3], indirect=True) # issue alice 3 creds
 async def test_get_credential_record_with_limit(
     alice_member_client: RichAsyncClient,
     issue_alice_many_creds: List[CredentialExchange],  # pylint: disable=unused-argument
@@ -80,9 +80,10 @@ async def test_get_credential_record_with_limit(
         {"limit": 3},
         {"limit": 4},
         {"limit": 1, "offset": 4},
+        {"limit": 2, "offset": 2}
     ]
 
-    expected_length = [1, 2, 3, 3, 0]
+    expected_length = [1, 2, 3, 3, 0, 1]
 
     for params, length in zip(valid_params, expected_length):
         response = (
--- a/app/tests/e2e/test_wallet_credentials.py
+++ b/app/tests/e2e/test_wallet_credentials.py
@@ -66,6 +66,11 @@ async def test_get_and_delete_credential_record(
 
 
 @pytest.mark.anyio
+@pytest.mark.skipif(
+    TestMode.regression_run in TestMode.fixture_params,
+    reason="Skipping due to regression run",
+)
+@pytest.mark.parametrize("issue_alice_many_creds",[3],indirect=True)
 async def test_get_credential_record_with_limit(
     alice_member_client: RichAsyncClient,
     issue_alice_creds_non_revoke: List[  # pylint: disable=unused-argument
--- a/app/tests/e2e/test_wallet_credentials.py
+++ b/app/tests/e2e/test_wallet_credentials.py
@@ -70,24 +70,23 @@ async def test_get_credential_record_with_limit(
         CredentialExchange
     ],
 ):
-    credentials = (await alice_member_client.get(WALLET_CREDENTIALS_PATH)).json()
 
-    assert len(credentials["results"]) == 10
-
-    credentials = (
+    credentials_1 = (
         await alice_member_client.get(f"{WALLET_CREDENTIALS_PATH}?limit=5")
     ).json()
 
-    assert len(credentials["results"]) == 5
+    assert len(credentials_1["results"]) == 5
 
-    credentials = (
-        await alice_member_client.get(f"{WALLET_CREDENTIALS_PATH}?limit=5&offset=8")
+    credentials_2 = (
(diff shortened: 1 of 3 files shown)


app/tests/fixtures/credentials.py
```diff
@@ -433,3 +433,79 @@ async def get_or_issue_regression_cred_valid(
         referent=valid_credential["referent"],
         cred_def_revocable=valid_credential["cred_def_id"],
     )
+
+
+@pytest.fixture(scope="function")
+async def issue_alice_creds_non_revoke(
```
ff137, line 439: let's call it `issue_alice_many_creds` instead

app/tests/fixtures/credentials.py
```diff
@@ -433,3 +433,79 @@ async def get_or_issue_regression_cred_valid(
...
+    faber_client: RichAsyncClient,
+    alice_member_client: RichAsyncClient,
+    credential_definition_id: str,
+    faber_and_alice_connection: FaberAliceConnect,
+) -> List[CredentialExchange]:
+
+    faber_conn_id = faber_and_alice_connection.faber_connection_id
+
+    faber_cred_ex_ids = []
+    for i in range(10):
```
ff137, line 449: can we parameterise the number of creds somehow? Ideally this should be configurable by the test calling the fixture. GPT should know but I think there's a way to pass params to fixtures from the test using it.

For our test we don't need 10 - that'll slow down things. Just 3 should be enough.

app/tests/e2e/test_wallet_credentials.py
```diff
@@ -60,3 +61,32 @@ async def test_get_and_delete_credential_record(
...
+
+    assert len(credentials_2["results"]) == 5
+    assert credentials_1 != credentials_2
+
+    credentials_3 = (
+        await alice_member_client.get(f"{WALLET_CREDENTIALS_PATH}?limit=5")
+    ).json()
+
+    assert len(credentials_3["results"]) == 5
+    assert credentials_1 == credentials_3
```
ff137, line 92: There's a few more cases to cover.

Let's say it's changed to just 3 credentials. We want to assert: limit 1,2,3 all return expected, and limit 4 returns 3.
Then, limit 1 + offset 3 should return an empty list.

For completeness we can also assert invalid params:
```py
        # Test invalid limit and offset values
        invalid_params = [
            {"limit": -1},  # must be positive
            {"offset": -1},  # must be positive
            {"limit": 0},  # must be greater than 0
            {"limit": 10001},  # must be less than or equal to max in ACA-Py: 10'000
        ]

        for params in invalid_params:
            with pytest.raises(HTTPException) as exc:
                await alice_member_client.get(BASE_PATH, params=params)
            assert exc.value.status_code == 422
```

Look to the `test_get_connections_paginated` for what assertions we currently do on paginated records

app/tests/fixtures/credentials.py
ff137, line 439: Test should probably be skipped in regression mode, to prevent bloating of regression fixtures.

ff137 (changes requested):

app/tests/fixtures/credentials.py
```diff
@@ -433,3 +433,80 @@ async def get_or_issue_regression_cred_valid(
...
+            filter_map={
+                "thread_id": thread_id,
+            },
+        )
+
+        cred_exchange = (
+            await alice_member_client.get(
+                f"{CREDENTIALS_BASE_PATH}?thread_id={thread_id}"
+            )
+        ).json()[0]
```
ff137, line 484: In general: optimise for readability. This is awaiting, calling .json(), and taking first element of list. Rather split it up:

```py
response = await alice_member_client.get(
                f"{CREDENTIALS_BASE_PATH}?thread_id={thread_id}"
            )
cred_exchange = response.json()[0]
```

And later:
```py
        await alice_member_client.post(
            f"{CREDENTIALS_BASE_PATH}/{cred_exchange['credential_exchange_id']}/request",
            json={},
        )
```
^ rather assign `cred_exchange['credential_exchange_id']` to a variable and use that. Again for readability - please always prefer doing it this way.
___

But here's a better way. Capture the webhook that was awaited by thread_id, and get alice's cred_ex_id from there:
```py
        alice_event = await check_webhook_state(
            client=alice_member_client,
            topic="credentials",
            state="offer-received",
            filter_map={
                "thread_id": thread_id,
            },
        )
        alice_cred_ex_id = alice_event ["credential_exchange_id"]
```

and then there's no need to fetch the record and do `cred_exchange['credential_exchange_id']` - can just directly do the `POST /request`

app/tests/e2e/test_wallet_credentials.py
```diff
@@ -60,3 +62,42 @@ async def test_get_and_delete_credential_record(
...
+async def test_get_credential_record_with_limit(
+    alice_member_client: RichAsyncClient,
+    issue_alice_many_creds: List[CredentialExchange],  # pylint: disable=unused-argument
+):
+    valid_params = [
+        {"limit": 1},
+        {"limit": 2},
+        {"limit": 3},
+        {"limit": 4},
+        {"limit": 1, "offset": 4},
```
ff137, line 82: Just add one more for completeness:
`{"limit": 2, "offset": 2}`

expected length for that = 1

app/tests/e2e/test_wallet_credentials.py
```diff
@@ -60,3 +62,42 @@ async def test_get_and_delete_credential_record(
...
         )
     assert exc.value.status_code == 404
+
+
+@pytest.mark.anyio
+@pytest.mark.skipif(
+    TestMode.regression_run in TestMode.fixture_params,
+    reason="Skipping due to regression run",
+)
+@pytest.mark.parametrize("issue_alice_many_creds", [3], indirect=True)
```
ff137, line 72: worth adding small comment: # issue alice 3 creds

ff137 (approved):
Just some minor changes requested, otherwise good 👍

PR #1171, opened 2024-11-08
:fire: Remove NATS kv

cl0ete:
Remove NATS kv from local and dev

### DIFF
--- a/tilt/cloudapi/nats/templates/job.yaml
+++ b/tilt/cloudapi/nats/templates/job.yaml
@@ -33,11 +33,6 @@ spec:
             - sh
             - -c
             - |
-              {{- range $kv := .Values.postInstall.kvs }}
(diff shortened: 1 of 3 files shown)


rblaine95 (approved):

tilt/cloudapi/nats/templates/job.yaml
```diff
@@ -33,11 +33,6 @@ spec:
             - sh
             - -c
             - |
-              {{- range $kv := .Values.postInstall.kvs }}
```
henrymsiska, line 36: suggest that the config remains as it will be used in future

tilt/cloudapi/nats/values.yaml
```diff
@@ -5,10 +5,6 @@ fullnameOverride: nats
 postInstall:
   enabled: true
 
-  kvs:
```
henrymsiska, line 8: suggest that this be configured as an empty list

henrymsiska (approved):
some minor suggestions then LGTM 🖖

PR #1194, opened 2024-11-21
:sparkles:Improve error detail when re-requesting credential

ff137:
Changes 400: `create_request() called multiple times for v2.0 credential exchange ...` 
to 409: `Credential ... has already been requested`

Edit: needs #1196 

Edit2: also needs custom wallet settings to not autocomplete flow, in order to replicate


cl0ete (approved):

ff137:
Cannot replicate the error

PR #1205, opened 2024-11-26
✅ Update test fixtures to did-exchange protocol

cl0ete:
Create did-exchange when needed.

Add work-around for finding connection without alias (Use alice did of connection)


app/tests/conftest.py
```diff
@@ -25,7 +25,7 @@
     trust_registry_client,
 )
 from app.tests.fixtures.member_connections import (
-    acme_and_alice_connection,
+    acme_and_alice_oob_connection,
```
ff137, line 28: can be renamed back, since oob is recommended way

app/tests/e2e/test_connections.py
```diff
@@ -172,18 +172,24 @@ async def test_get_connections(
...
 
 @pytest.mark.anyio
 async def test_get_connection_by_id(
     bob_member_client: RichAsyncClient,
+    alice_member_client: RichAsyncClient,
 ):
-    invitation_response = await bob_member_client.post(f"{BASE_PATH}/create-invitation")
-    invitation = invitation_response.json()
-    connection_id = invitation["connection_id"]
+    connection_alias = "TempAliceBobConnectionById"
```
ff137, line 182: We may as well add an assertion at the end, that the fetched record has this alias. Doesn't seem like alias is used otherwise.

Also, worth noting this test is run for regression tests too, in which case alice+bob members will keep gaining new records. We should therefore add a delete call for this connection record at the end of the test. Best way to do it is:

```py
# as it is currently, establish connection:
    bob_and_alice_connection = await create_bob_alice_connection(
        alice_member_client, bob_member_client, alias=connection_alias
    )

    bob_connection_id = bob_and_alice_connection.bob_connection_id

#Then start try block:
try:
... # get and assertions
finally:
... # call delete endpoint for this record
```

So that if anything goes wrong for whatever reason, the delete will always be called after connection created.

app/tests/e2e/test_connections.py
```diff
@@ -195,56 +201,22 @@ async def test_get_connection_by_id(
...
-        await alice_member_client.get(f"{BASE_PATH}/{alice_connection_id}")
-    ).json()
+    response = await bob_member_client.delete(f"{BASE_PATH}/{bob_connection_id}")
+    assert response.status_code == 204
 
-    assert "completed" in alice_connection["state"]
-    assert "completed" in bob_connection["state"]
+    with pytest.raises(HTTPException) as exc:
+        response = await bob_member_client.get(f"{BASE_PATH}/{bob_connection_id}")
+    assert exc.value.status_code == 404
```
ff137, line 219: We used to delete for bob and alice. Maybe just add some comments that record is now deleted for Alice too. Then, to validate that, have a 404 assert if alice tries to delete her connection after bob has deleted.

app/tests/util/connections.py
```diff
@@ -260,3 +299,44 @@ async def fetch_or_create_trust_registry_connection(
...
+
+
+async def create_did_exchange(
+    bob_member_client: RichAsyncClient, alice_member_client: RichAsyncClient, alias: str
+) -> BobAliceConnect:
+
+    # Get Bob's public DID
+    bob_public_did = (await bob_member_client.get(f"{DID_BASE_PATH}/public")).json()[
+        "did"
+    ]
```
ff137, line 311: Bad style - it's an await, .json, and reading key all in one line. Do await and then .json()[key] separately

It's mainly that doing `(await ..).method()` looks bad, but if it's _just_ (await ..).json() then it's fine, since it happens so often. Chaining more operations in one line after that it starts being hard to read

app/tests/util/connections.py
```diff
@@ -152,30 +169,49 @@ async def fetch_or_create_connection(
...
-            alice_member_client, bob_member_client, alias=alias
-        )
+        if did_exchange:
+            return await create_did_exchange(
+                bob_member_client=bob_member_client,
+                alice_member_client=alice_member_client,
+                alias=alias,
+            )
+        else:
+
```
ff137, line 201: remove empty line

app/tests/util/connections.py
```diff
@@ -63,28 +65,35 @@ async def assert_both_connections_ready(
...
         )
     ).json()
 
-    bob_connection_id = invitation["connection_id"]
-    alice_connection_id = invitation_response["connection_id"]
+    bob_connection = (
+        await bob_member_client.get(
+            f"{CONNECTIONS_BASE_PATH}?alias={alias}",
+        )
+    ).json()[0]
```
ff137, line 93: Hmm ... taking 0th element of list can have unexpected behavior. Either wait for webhook and read connection_id from that, or add another query param filter, like their_did, using alice's my_did, if that's possible.

app/tests/util/connections.py
ff137: P.S. Add a note that bob is assumed to have a public did, so "bob" is an issuer in this context

ff137 (changes requested):

ff137 (approved):

PR #1206, opened 2024-11-27
✨ Add endpoint to fix revocation registry state

cl0ete:
Add fix-rev-reg endpoint: `/v1/issuer/credentials/fix-revocation-registry/{revocation_registry_id}`

### DIFF
--- a/app/routes/revocation.py
+++ b/app/routes/revocation.py
@@ -340,7 +340,8 @@ async def fix_revocation_registry_entry_state(
     Query Parameters:
     ---
         apply_ledger_update: bool
-            Set to True to apply the ledger update
+            Set to True to apply the ledger update, only calculates the difference
+            between the wallet and ledger state if False
 
     Returns:
     ---
--- a/app/routes/revocation.py
+++ b/app/routes/revocation.py
@@ -1,11 +1,12 @@
 import asyncio
 from typing import Optional
 
-from aries_cloudcontroller import IssuerCredRevRecord
+from aries_cloudcontroller import IssuerCredRevRecord, RevRegWalletUpdatedResult
 from fastapi import APIRouter, Depends
 
 from app.dependencies.acapy_clients import client_from_auth
 from app.dependencies.auth import AcaPyAuth, acapy_auth_from_header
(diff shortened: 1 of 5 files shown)


app/routes/revocation.py
```diff
@@ -313,3 +313,50 @@ async def get_pending_revocations(
...
+    Query Parameters:
+    ---
+        apply_ledger_update: bool
+            Set to True to apply the ledger update
+
+    Returns:
+    ---
+        RevRegWalletUpdatedResult
+    """
+    bound_logger = logger.bind(body={"revocation_registry_id": revocation_registry_id})
```
ff137, line 349: include `apply_ledger_update` to `body={..}`

app/routes/revocation.py
```diff
@@ -313,3 +313,50 @@ async def get_pending_revocations(
...
+
+    Path Parameters:
+    ---
+        revocation_registry_id: str
+            The ID of the revocation registry for which to fix the state
+
+    Query Parameters:
+    ---
+        apply_ledger_update: bool
+            Set to True to apply the ledger update
```
ff137, line 343: What does False do? Would be good to just add a note, like False applies to wallet state, but not ledger (idk the specifics)

app/tests/e2e/test_revocation.py
```diff
@@ -347,3 +347,30 @@ async def test_get_pending_revocations_bad_payload(
...
+        ),
+        ("bad_format", 422),
+    ],
+)
+async def test_fix_rev_reg_bad_id(
+    faber_client: RichAsyncClient, rev_reg_id: str, status_code: int
+):
+    with pytest.raises(HTTPException) as exc:
+        await faber_client.put(
+            f"{CREDENTIALS_BASE_PATH}/fix-revocation-registry/{rev_reg_id}",
```
ff137, line 372: Nitpick but may as well change: these tests should use REVOCATION_BASE_PATH, instead of credential base path, now that it's moved. Right now it's the same. May as well refactor.

ff137 (approved):
Nice - just some minor requests

PR #1217, opened 2024-12-04
⚡️Waypoint re-subscribe

cl0ete:
Add retry logic to subscribe function.
Also resubscribe on TimeoutError. 

Update unit tests


waypoint/services/nats_service.py
```diff
@@ -27,42 +29,53 @@ def __init__(self, jetstream: JetStreamContext):
...
-            logger.error("Error subscribing to NATS: {}", e)
-            raise
+        # Format the time in the required format
+        start_time = look_back_time.strftime("%Y-%m-%dT%H:%M:%S.%f")[:-3] + "Z"
+        config = ConsumerConfig(
+            deliver_policy=DeliverPolicy.BY_START_TIME,
+            opt_start_time=start_time,
+        )
+
+        @retry(TimeoutError, delay=1, backoff=2, max_delay=16, logger=logger)
```
ff137, line 58: I first recommended `retry`, but let's look into using `tenacity` instead. `retry` should work, but is unmaintained for many years now.

`tenacity` example a la GPT:
```py
from tenacity import retry, stop_after_attempt, wait_exponential

@retry(stop=stop_after_attempt(5), wait=wait_exponential(multiplier=1, min=1, max=10))
def unstable_function():
    print("Trying...")
    raise Exception("Failure")

unstable_function()
```

waypoint/services/nats_service.py
```diff
@@ -106,14 +119,20 @@ async def event_generator():
...
+                        logger.trace("Received event: {}", event)
                         yield CloudApiWebhookEventGeneric(**event)
                         await message.ack()
-                except TimeoutError:
+                except FetchTimeoutError:
                     logger.trace("Timeout fetching messages continuing...")
                     await asyncio.sleep(0.1)
+                except TimeoutError:
+                    logger.error("FATAL: Timeout Error")
+                    raise
```
ff137, line 135: I assume this only ends up being raised when retry logic fails too many times? Perhaps it's worth using `critical` log level, if it's appropriate. And maybe just print the exception string with the log

waypoint/services/nats_service.py
ff137: See: https://github.com/jd/tenacity

waypoint/services/nats_service.py
```diff
@@ -15,6 +18,15 @@
...
 
+def retry_log(retry_state: RetryCallState):
+    """Custom logging for retry attempts."""
+    if retry_state.outcome.failed:
+        exception = retry_state.outcome.exception()
+        logger.warning(
+            f"Retry attempt {retry_state.attempt_number} failed due to {type(exception).__name__}: {exception}"
+        )
+
+
```
ff137, line 29: This can be made a private method of the NatsEventProcessor class

Also the log line can use the {}-interpolation pattern

waypoint/services/nats_service.py
```diff
@@ -25,44 +37,63 @@ def __init__(self, jetstream: JetStreamContext):
...
+            deliver_policy=DeliverPolicy.BY_START_TIME,
+            opt_start_time=start_time,
+        )
 
-            return subscription
+        @tenacity.retry(
+            retry=tenacity.retry_if_exception_type(TimeoutError),
+            wait=tenacity.wait_exponential(multiplier=1, max=16),
+            after=retry_log,
+            stop=tenacity.stop_never,
```
ff137, line 70: just for neatness, we can import all these `tenacity.*` members from the package, so we can reference them without the `tenacity.` 

Also perhaps add a comment to summarise: "this decorator will retry indefinitely if TimeoutError occurs, with exponential back-off between retries". Something like that

waypoint/services/nats_service.py
```diff
@@ -25,44 +37,63 @@ def __init__(self, jetstream: JetStreamContext):
...
-        except BadSubscriptionError as e:
-            logger.error("BadSubscriptionError subscribing to NATS: {}", e)
-            raise
-        except Error as e:
-            logger.error("Error subscribing to NATS: {}", e)
-            raise
+        try:
+            return await pull_subscribe(config, **subscribe_kwargs)
         except Exception:
             logger.exception("Unknown error subscribing to NATS")
```
ff137, line 98: Btw this string doesn't have to say "Unknown error", it can now just be An exception occurred

waypoint/services/nats_service.py
```diff
@@ -86,44 +117,112 @@ async def process_events(
...
+
+        async def event_generator(
+            *,
+            subscription: JetStreamContext.PullSubscription,
+            group_id: str,
+            wallet_id: str,
+            topic: str,
+            state: str,
+            stop_event: asyncio.Event,
+            start_time: str,
```
ff137, line 137: the `start_time` arg will mask the `start_time` that was just defined in line 127

if start_time is never modified again, then it doesn't have to be passed as an arg, as it'll be set in the method definition

waypoint/services/nats_service.py
```diff
@@ -86,44 +117,112 @@ async def process_events(
...
+                    remaining_time = end_time - time.time()
+                    logger.trace("remaining_time: {}", remaining_time)
+                    if remaining_time <= 0:
+                        logger.debug("Timeout reached")
+                        stop_event.set()
+                        break
+
+                    try:
+                        messages = await subscription.fetch(
+                            batch=5, timeout=0.2, heartbeat=0.02
```
ff137, line 151: I think a heartbeat of 50 times a second is overkill. And maybe timeout can be relaxed a lil too. 

I propose timeout can be 0.5, and heartbeat 0.2. But idk, up to you + Willem

waypoint/services/nats_service.py
```diff
@@ -86,44 +117,112 @@ async def process_events(
...
+                subscription=subscription,
+                stop_event=stop_event,
+                group_id=group_id,
+                wallet_id=wallet_id,
+                topic=topic,
+                state=state,
+                start_time=start_time,
+            )
+        except Exception as e:  # pylint: disable=W0718
+            logger.exception("Unexpected error processing events: {}", e)
```
ff137, line 213: the e is not necessary, as .exception automatically prints it

ff137:
Remember to update PR emoji/title when it's ready to review / merge :-)

waypoint/services/nats_service.py
```diff
@@ -184,7 +193,7 @@ async def event_generator(
                         logger.debug("Successfully resubscribed to NATS.")
 
                     except Exception as e:  # pylint: disable=W0718
-                        logger.exception("Unexpected error in event generator: {}", e)
+                        logger.exception("Unexpected error in event generator: {}")
```
ff137, line 195: ~`: {}`~

ff137 (approved):
Nice! Just the one line to be edited

PR #1263, opened 2025-01-15
:art: Clean up waypoint service and look_back defaults

ff137:
~~Adds a `durable_name` config to the waypoint NATS subscription, so that if an event gets acked, it can still be re-consumed in a subsequent request.~~

~~This means clients can receive the same event again when they re-connect. This may be desirable, or not. If not, it can be controlled using the `look_back` parameter. So overall, I think it's a good addition.~~
___
Reverted the above, because nats.py doesn't work as expected when testing with durable_name config. It may already re-fetch already acked messages when resubscribing ... to be confirmed
___

:art: Also does some cleanup of method args (some cases of method names shadowing vars that already exist in scope), and ensuring that SSE_LOOK_BACK default is used properly.

:test_tube: Increases the default look_back in tests from 5 to 15. (Seems to be some tests that sporadically fail with the shorter look_back.) Also skips a regression test, due to sporadic failure; pagination doesn't need to be validated in regression mode.

:heavy_plus_sign: Adds `uuid_utils` as dependency


cl0ete:
The stream that waypoint subs to has a TTL of 1 min

cl0ete:
> Reverted the above, because nats.py doesn't work as expected when testing with durable_name config. It may already re-fetch already acked messages when resubscribing ... to be confirmed

Yes as I understand it, it will re-fetch messages as it is now. With each sub an ephemeral consumer (NATS terminology) will be created that has not acked on any messages.

cl0ete (approved):

PR #1269, opened 2025-01-20
✨ Support using podman & improve tilt env reuse

henrymsiska:
✨ Support tilt running in podman (with docker compatibility mode):
  * switched to using tilt's `custm_build` instead of `docker_build` with `skips_local_docker=True` and pushing to local repository
  * introiduced `
DOCKER_HOST = "unix:///var/run/docker.sock"` env var pointing at default docker socket location

🔧 Adds support for mise test task to run e2e tests and provide parameters to the test tasks, e.g:
  * `mise run tests -- -n 4`
  * `mise run tests:e2e -- -n 4`
  
🔧 mise kind, registry and cache tasks fixes:
* make the docker cache and registry container names configurable via env
* fix destroy and destroy all mise scripts to handle docker cache and registry correctly and reuse running/existing to allow sharing of these when working with multiple projects

🔧 adds mise  tasks to format source code
* `mise run fmt`
* `misse run fmt:check`


ff137 (commented):
LGTM :+1: 

Just need to get the workflows passing -- docker/login-action is failing in tests
https://github.com/didx-xyz/aries-cloudapi-python/actions/runs/12862232674/job/35856858278?pr=1269

cl0ete (approved):
Pulled branch and ran tests & regression tests locally, all passed

.mise.toml
```diff
@@ -81,10 +91,47 @@ run = """
...
+"""
+
+[tasks."fmt:tiltfiles"]
+description = "Format Tiltfiles"
+run = """
+#!/bin/bash
+find . -type f -name "Tiltfile" | while read -r file; do
+    black "$file"
+  done
+"""
```
rblaine95, line 137: I think it'd be better to merge the format tasks:

```suggestion
[tasks.fmt]
description = "Format files"
run = """
#!/bin/bash
isort . --profile black --diff
black . --diff
find . -type f -name "Tiltfile" | while read -r file; do
    black "$file"
  done
"""

[tasks."fmt:check"]
description = "Check all files for formatting"
run = """
#!/bin/bash
isort . --check --profile black --diff
black . --check --diff
find . -type f -name "Tiltfile" | while read -r file; do
    black --check --diff "$file"
  done
"""
```

rblaine95 (changes requested):

rblaine95 (approved):

PR #1293, opened 2025-01-27
⚡️ Cache schema_Id calls from ledger

cl0ete:
Added `aiocache` to `app/`

Added constant `CACHE_TTL` to `shared/constants` default 60

Use `cred_def_id` as cache key


ff137 (commented):
I think we can make maxsize much greater.

The question I have is how the caching works with multiple controllers.

Does maxsize=6 mean it caches 6 unique combinations of controller + cred def arguments?

In which case, maxsize of 1000+ could be appropriate

cl0ete:
> I think we can make maxsize much greater.
> 
> The question I have is how the caching works with multiple controllers.
> 
> Does maxsize=6 mean it caches 6 unique combinations of controller + cred def arguments?
> 
> In which case, maxsize of 1000+ could be appropriate

I jumped the gun a bit. lru_cache does not support async calls. Looking at [aiocache](https://github.com/aio-libs/aiocache)

app/services/acapy_ledger.py
```diff
@@ -153,6 +155,8 @@ async def accept_taa_if_required(aries_controller: AcaPyClient) -> None:
         )
 
 
+# Grab cred_def_id from args to use as cache-key
+@cached(ttl=CACHE_TTL, cache=SimpleMemoryCache, key_builder=lambda *args: args[1])
```
ff137, line 159: We can probably cache indefinitely - i.e. no TTL - since schema_id will always be the same for a given cred def.

ff137 (commented):
P.S. remember to merge main, so the regression tests won't break randomly

app/services/acapy_ledger.py
```diff
@@ -153,6 +154,8 @@ async def accept_taa_if_required(aries_controller: AcaPyClient) -> None:
         )
 
 
+# Grab cred_def_id from args to use as cache-key
+@cached(cache=SimpleMemoryCache, key_builder=lambda *args: args[1])
```
ff137, line 158: I wrote a test for this -- it looks like it should be args[2] to successfully capture the cred_def_id from the args

app/tests/services/test_acapy_ledger.py
```diff
@@ -102,3 +102,83 @@ async def test_schema_id_from_credential_definition_id_schema_id(
...
+    )
+
+    await schema_id_from_credential_definition_id(
+        mock_agent_controller, cred_def_id_old2
+    )
+
+    # Assert both API calls were made
+    verify(mock_agent_controller.schema).get_schema(schema_id="456")
+    verify(mock_agent_controller.schema).get_schema(schema_id="789")
+    assert False
```
ff137, line 184: 👀

ff137 (approved):

PR #1297, opened 2025-01-27
:arrow_down: Downgrade DDTrace

rblaine95:
* There seems to be an issue in `2.20`
* Downgrade to `2.19` to verify


rblaine95:
Looks like we lost APM on 17 January, when we upgraded to Python 3.13 🤔

PR #1299, opened 2025-01-28
:heavy_minus_sign: Remove DDTrace

rblaine95:
* We might not need to install it as we're using Auto Instrumentation to inject it in K8s

---

Did some testing, confirmed, we don't need to install DDTrace and can just use the Auto-Instrumentation.
Continuous Profiler and APM works.


ff137 (approved):
Thanks for making the deps alphabetical too, my OCD is happy 😄

Does this mean we can bump back to 3.13, or is it still a blocker for what version can be injected?

rblaine95:
> Does this mean we can bump back to 3.13, or is it still a blocker for what version can be injected?

Unfortunately no, it doesn't seem the auto-instrumentation currently supports Python 3.13 😞 
That's where the "unsupported runtime detected" warning/error comes from

PR #1309, opened 2025-02-03
:adhesive_bandage: Increase `retry_delay` in wait for issuer onboard

ff137:
And :wrench: make configurable

🧪 See if this relaxes acapy load just a lil during many issuer onboards / help fix sporadic 504 errors


ff137:
No longer needed

PR #1311, opened 2025-02-04
♻️👷 Refactor CICD Pipeline

rblaine95:
* Merge `style-check.yml` into CICD
* Major refactor of CICD pipeline
  * Use composite actions to make CICD pipeline more modular and maintainable
* Currently the OSS deployment is 1 replica of everything
  * Except Ledger Nodes as that requires 4

---

Deploy and Test EKS is now a single job consisting of Composite actions.
This makes maintaining the CICD pipeline easier as it modularizes the workflow a bit.
This also means that there global concurrency group isn't on the entire pipeline, but instead just on the Deploy and Test EKS.
This means that, other than Deploy and Test EKS, multiple workflows can run simultaneously.

Previously, for running the local tests, we used a matrix to create 5 Github hosted runners to run 5 copies of the local dev stack, and then run subsets of the tests. We would then combine the test coverage reports.
Now, due to optimizations made with parallelizing tests and to the local dev infrastructure, we can run everything on a single Github hosted runner, greatly simplifying the local testing and coverage reporting.

Closes DIDC-105 and DIDC-305

---

FYI, @ff137 and @cl0ete, due to the removal of `uv` you will need to reinitialize the `.venv` (`rm -rf .venv && mise run poetry:install`)


ff137:
PS: there is an old CICD readme that was started almost 2 years ago 👀 `.github/workflows/README.md`

It's linked to under the "CI/CD" subheading on the main readme. It might be worth giving that a lil comb over. Doesn't have to be extensive, or done in this PR. Just wanted to give a heads up

cl0ete (approved):
LGTM :eyes:

helm/acapy-test/conf/dev/values.yaml
```diff
@@ -0,0 +1,43 @@
+command:
+  - sh
+  - -c
+  - |
+    poetry run pytest \
+      -n 4 \
```
ff137, line 6: this can maybe be made -n 6, because we mark 5 xdist groups in our tests. So 4 would limit concurrency a lil

.github/workflows/cicd.yml
```diff
@@ -0,0 +1,596 @@
...
+          poetry run pylint \
+            app/ \
+            endorser/ \
+            shared/ \
+            trustregistry/ \
+            waypoint/ \
+            --rcfile=.pylintrc \
+            -r n \
+            --msg-template="{path}:{line}: [{msg_id}({symbol}), {obj}] {msg}" \
+            --exit-zero | tee pylintreport.txt
```
ff137, line 167: Is the pylintreport used anywhere, like for submitting to codacy/sonarcloud?

Because if not, I think we can maybe remove it.

Plan is to eventually switch over to using ruff instead, so black/isort/pylint is all included in ruff. 

And for the time being, this pylint run will mainly just add ~40 seconds. I don't think it'll ever error, it just reports code quality, which we already get from codacy/sonarcloud

ff137 (approved):
Fantastic! Awesome work

Mainly just a comment about the pylint job maybe being removable

.github/workflows/cicd.yml
rblaine95: Yeah, I was wondering about it, it doesn't seem to report to anywhere.
Will remove in next commit 👍

rblaine95: @ff137 should I also remove the `.pylintrc` file?

helm/acapy-test/conf/dev/values.yaml
rblaine95: I've increased the jobs from 4 to 6, let's see if it speeds things up at all 🤔

ff137: Pretty much the same

.github/workflows/cicd.yml
ff137: We can keep that for now -- will maybe assist in ruff config, and can delete after switching over

PR #1316, opened 2025-02-07
:construction_worker: 🦨 Run local tests on Blacksmith

rblaine95:


rblaine95:
Seems like Blacksmith speeds up the testing considerably but the startup time is unaffected.
So in total Blacksmith speeds up Local Tests by ±60s

rblaine95:
It's not really worth it 🤷 
We're not particularly CPU bound, mainly IO

PR #1321, opened 2025-02-10
:construction_worker: Deduplicate Agent Images

rblaine95:
* Governance and Multitenant Agents are technically the exact same
* Only difference is the `acapy-wallet-groups-plugin` and startup parameters
* Update Mediator `py3.12-1.2.1` -> `py3.12-1.2.2`


ff137 (approved):
Very nice

cl0ete (approved):

PR #1331, opened 2025-02-14
🐛 Waypoint wait for consumer to be ready

cl0ete:
:wrench: Adds new env vars:
```py
BATCH_SIZE = int(os.getenv("NATS_BATCH_SIZE", "5"))  # number of messages to get per fetch request
TIMEOUT = float(os.getenv("NATS_TIMEOUT", "0.5"))  # fetch timeout
HEARTBEAT = float(os.getenv("NATS_HEARTBEAT", "0.1"))  # fetch heartbeat
MAX_TIMEOUT_ERRORS = int(os.getenv("NATS_MAX_TIMEOUT_ERRORS", "3"))  # number of heartbeat timeouts before reconnecting
```

➕ Adds uuid-utils to waypoint - to generate and add a UUID to waypoint requests, for better log tracing

✅ 99% test coverage for NATS processor


waypoint/services/nats_service.py
```diff
@@ -93,6 +93,18 @@ async def pull_subscribe():
...
+                not_ready = True
+                while not_ready:
+                    consumer_info = await subscription.consumer_info()
+                    if isinstance(consumer_info, ConsumerInfo):
+                        bound_logger.trace(
+                            "Consumer is ready {}, {}",
+                            consumer_info.name,
+                            consumer_info.stream_name,
+                        )
+                        not_ready = False
```
ff137, line 106: I can't figure out from the nats.py code if this actually ensures the consumer is ready. ConsumerInfo has:
```py
@dataclass
class ConsumerInfo(Base):
    """
    ConsumerInfo represents the info about the consumer.
    """

    name: str
    stream_name: str
    config: ConsumerConfig
    # FIXME: Do not handle dates for now.
    # created: datetime
    delivered: Optional[SequenceInfo] = None
    ack_floor: Optional[SequenceInfo] = None
    num_ack_pending: Optional[int] = None
    num_redelivered: Optional[int] = None
    num_waiting: Optional[int] = None
    num_pending: Optional[int] = None
    cluster: Optional[ClusterInfo] = None
    push_bound: Optional[bool] = None
```
and none of it stands out as something we can check that it's ready. So we just assume if we get this response, it's ready. Only way to confirm that is if it fixes the bug ... so, let me know if it does

One nitpick is that I prefer: `ready = False` and `while not ready`, then in loop `ready = True`.
I think the if isinstance check can be dropped.

___
Come to think about it, the while loop is probably redundant, because the await can only return ConsumerInfo, or raise exception. So I don't think it'll ever loop. 

So, presumably we just wanna get consumer info before proceeding.

waypoint/services/nats_service.py
```diff
@@ -93,6 +96,15 @@ async def pull_subscribe():
                 subscription = await self.js_context.pull_subscribe(
                     config=config, **subscribe_kwargs
                 )
+
+                consumer_info = await subscription.consumer_info()
+                if isinstance(consumer_info, ConsumerInfo):
```
ff137, line 101: If isinstance check isn't necessary, it'll always return that type

cl0ete: Yes. I don't even think getting the info helps with the issue we were trying to solve... Think the llms were wrong...

PR #1338, opened 2025-02-27
:construction_worker: Notify didx-cloud

rblaine95:
* Don't trigger CD directly
* Rather notify
* Notify will then handle triggering CD
* Also forcefully delete the pod that fetches the EKS test results
  * Speed things up a bit


henrymsiska (approved):
🫡

cl0ete (approved):

PR #1351, opened 2025-03-03
:construction_worker: Fix PR Length Check

rblaine95:
* Add global to emoji regex
* Trim trailing spaces in PR title
* Add some debug output


cl0ete (approved):

ff137 (commented):
PS. When there is no emoji in the title, the sample comment prints: use 🔥 or :fire: -- the second one should be wrapped in `` block

PR #1354, opened 2025-03-03
✨1 Anoncreds schemas

cl0ete:
Add anoncreds schema support

:boom: Removes `schema_id` query parameter from "get_schemas" endpoint (`GET /schemas`) -- since there is already a dedicated endpoint for fetching by schema_id (`GET /schemas/{schema_id}`) -- and the query param was supported in acapy for fetching indy schemas, but no longer for anoncreds schemas, so we adapt and remove it.

### DIFF
--- a/app/models/definitions.py
+++ b/app/models/definitions.py
@@ -29,7 +29,10 @@ class CredentialDefinition(BaseModel):
 
 
 class CreateSchema(BaseModel):
-    schema_type: SchemaType = SchemaType.INDY
+    schema_type: SchemaType = Field(
+        default=SchemaType.INDY,
+        description="The type of schema to create. Supported values are 'indy' and 'anoncreds'.",
+    )
     name: str = Field(..., examples=[sample_name])
     version: str = Field(..., examples=[sample_version])
     attribute_names: List[str] = Field(..., examples=[sample_attribute_names])
--- a/app/models/definitions.py
+++ b/app/models/definitions.py
@@ -29,7 +29,7 @@ class CredentialDefinition(BaseModel):
 
 
 class CreateSchema(BaseModel):
-    schema_type: SchemaType = SchemaType.ANONCREDS
+    schema_type: SchemaType = SchemaType.INDY
     name: str = Field(..., examples=[sample_name])
     version: str = Field(..., examples=[sample_version])
     attribute_names: List[str] = Field(..., examples=[sample_attribute_names])
--- a/app/routes/definitions.py
+++ b/app/routes/definitions.py
@@ -136,7 +136,6 @@ async def get_schemas(
                 schema_name=schema_name,
                 schema_version=schema_version,
             )
-
         else:  # Governance is calling the endpoint
             try:
                 schemas = await schemas_service.get_schemas_as_governance(
@@ -190,11 +189,9 @@ async def get_schema(
 
     async with client_from_auth(auth) as aries_controller:
         if is_governance:
-            # controller.settings.get_settings() returns None ????
             # Get the wallet type from the server config
             server_config = await aries_controller.server.get_config()
             wallet_type = server_config.config.get("wallet.type")
-
         else:
             wallet_type = await get_wallet_type(
                 aries_controller=aries_controller,
@@ -202,7 +199,6 @@ async def get_schema(
             )
 
         if wallet_type == "askar-anoncreds":
-
             schema = await handle_acapy_call(
                 logger=bound_logger,
                 acapy_call=aries_controller.anoncreds_schemas.get_schema,
@@ -213,9 +209,7 @@ async def get_schema(
                 raise HTTPException(404, f"Schema with id {schema_id} not found.")
 
             result = anoncreds_schema_from_acapy(schema)
-
         elif wallet_type == "askar":
-
             schema = await handle_acapy_call(
                 logger=bound_logger,
                 acapy_call=aries_controller.schema.get_schema,
@@ -226,8 +220,8 @@ async def get_schema(
                 raise HTTPException(404, f"Schema with id {schema_id} not found.")
 
             result = credential_schema_from_acapy(schema.var_schema)
-
         else:
+            # Should never happen
             raise HTTPException(500, "Unknown wallet type")
 
     bound_logger.debug("Successfully fetched schema by id.")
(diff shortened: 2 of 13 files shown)


ff137 (commented):
The main goal here will be:
- support both of the acapy endpoints for indy + anoncreds 
- we can deduplicate the models for those two cases, and accept a generic request model for schemas, perhaps with a "type" field for switching between indy and anoncreds
  - and generic (deduplicated) response model - trying to take the overlap of the two
- if we can't get acapy to support the indy endpoints with an anoncreds wallet, then we just enforce that "type: anoncreds" can only be used with anoncreds wallet, and same with askar wallet-type + indy

So that's the goal to have an interface to switch between existing methods and the new anoncreds one.

And then for testing we can ideally just parameterise over "type", to reuse existing tests and cover anoncreds cases too

app/util/tenants.py
```diff
@@ -122,3 +122,28 @@ def assert_valid_group(
...
+    controller_wallet_id = get_wallet_id_from_b64encoded_jwt(controller_token)
+    async with get_tenant_admin_controller() as admin_controller:
+        wallet = await handle_acapy_call(
+            acapy_call=admin_controller.multitenancy.get_wallet,
+            wallet_id=controller_wallet_id,
+            logger=logger,
+        )
+        if not wallet:
+            logger.info("Bad request: Wallet not found.")
+            raise WalletNotFoundException(wallet_id=controller_wallet_id)
```
ff137, line 146: In this scenario, we should probably raise a 401 Unauthorized - since we're getting a request with a bad auth token.

app/util/definitions.py
```diff
@@ -13,6 +22,15 @@ def credential_schema_from_acapy(schema: ModelSchema):
     )
 
 
+def anon_schema_from_acapy(schema: GetSchemaResult):
```
ff137, line 25: just for consistency - prefer the full name "anoncreds" to "anon"

app/util/definitions.py
```diff
@@ -1,9 +1,18 @@
 from aries_cloudcontroller import CredentialDefinition as AcaPyCredentialDefinition
-from aries_cloudcontroller import ModelSchema
+from aries_cloudcontroller import GetSchemaResult, ModelSchema, SchemaState
 
 from app.models.definitions import CredentialDefinition, CredentialSchema
 
 
+def credential_anon_schema_from_acapy(schema: SchemaState):
```
ff137, line 7: perhaps "anoncreds_credential_schema"

app/services/definitions/schemas.py
```diff
@@ -149,38 +219,73 @@ async def get_schemas_as_governance(
...
+
+        # Wait for completion of futures
+        if get_schema_futures:
+            logger.debug("Fetching each of the created schemas")
+            schema_results: List[SchemaGetResult] = await asyncio.gather(
+                *get_schema_futures
+            )
+        else:
+            logger.debug("No created schema ids returned")
+            schema_results = []
```
ff137, line 277: The "wait for schema futures" can maybe be deduplicated -- moved after the wallet-type if-else's

app/services/definitions/schemas.py
```diff
@@ -71,6 +115,13 @@ async def get_schemas_as_tenant(
         }
     )
     bound_logger.debug("Fetching schemas from trust registry")
+    if await is_anoncreds_wallet(
+        aries_controller=aries_controller,
+        logger=bound_logger,
+    ):
+        wallet_type = "askar-anoncreds"
+    else:
+        wallet_type = "askar"
```
ff137, line 124: We can maybe split `is_anoncreds_wallet` into two functions -- one for fetching the wallet_type, and one for asserting it equals anoncreds. Because in this case, we'd just want to use `wallet_type = await get_wallet_type(..)`

app/services/definitions/schemas.py
```diff
@@ -38,15 +46,51 @@ async def create_schema(
...
+        )
+
+        result = await publisher.publish_schema(schema_request)
+
+    else:
+        raise CloudApiException(
+            # TODO: Improve message
+            "Wallet type does not match schema type. Cannot create schema.",
+            status_code=400,
+        )
```
ff137, line 93: Can maybe have some conditional formatted string, like
f"{schema.schema_type} can only be created by '{"askar" if schema.schema_type == "INDY" else "askar-anoncreds"}' wallet types"
Something like that.

Can asign schema_type to a var, and then assign "required_wallet_type" to  '{"askar" if schema.schema_type == "INDY" else "askar-anoncreds"}, and then just have "`if wallet_type != required_wallet_type, raise 400 with error message". Otherwise proceed with if schema_type check and doing anoncreds/indy call

app/routes/definitions.py
```diff
@@ -85,7 +87,6 @@ async def create_schema(
     response_model=List[CredentialSchema],
 )
 async def get_schemas(
-    schema_id: Optional[str] = None,
```
ff137, line 88: So, it's just anoncreds that doesn't support fetching by schema_id? We can maybe just make a note of that and ask if it can be implemented in acapy

app/models/definitions.py
```diff
@@ -23,6 +29,7 @@ class CredentialDefinition(BaseModel):
 
 
 class CreateSchema(BaseModel):
+    schema_type: SchemaType = SchemaType.ANONCREDS
```
ff137, line 32: I'd suggest to keep this default to SchemaType.INDY, for existing tests, so they don't require explicit changes. Anoncreds tests can then explicitly request anoncreds. Can change defaults after migration is successful.

app/util/tenants.py
cl0ete: :thinking:  Will a user with a bad token even get this far?

app/services/definitions/schemas.py
cl0ete: The types on these are slightly different so I just decided to keep wallet specific code in the if blocks.

app/routes/definitions.py
ff137: I see it doesn't make sense to fetch "all schemas" by a schema_id - and we already have an endpoint for fetching by schema_id. So I just noted it as a breaking change that we're removing the query param (not a big deal)

ff137 (approved):

PR #1379, opened 2025-03-12
✅ K6 didexchange

wdbasson:


ff137:
@wdbasson @cl0ete Please resolve SonarCloud security hotspots before merging. If they're safe / false positives, just mark them as accepted on the SonarCloud page. (Code duplication is fine, but there was 1 security hotspot that I marked as accepted.)

It adds to the "security issues" on the repo, and/or downgrades our code quality rating if we don't

PR #1382, opened 2025-03-13
:sparkles: Allow anoncreds issuers to create schemas

ff137:
Temporarily removes the restriction that only governance can create schemas.

This is to accommodate for the fact that we want to test both indy and anoncreds for the time being, and we cannot configure governance to be able to create both. So, we will allow anoncreds issuers to create schemas, until indy support is dropped.

- Removes duplicate auth from schema service
- Updates tests

### DIFF
--- a/app/routes/definitions.py
+++ b/app/routes/definitions.py
@@ -85,6 +85,7 @@ async def create_schema(
         )
         raise CloudApiException("Unauthorized", 403)
 
+    public_did = None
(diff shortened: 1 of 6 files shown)


cl0ete (commented):
Just check if the wallet type gotten in `app/services/definitions/schemas.py` in `create_schema` makes sense for the issuer case

cl0ete (approved):

PR #1389, opened 2025-03-17
:sparkles: Wait for anoncreds schema endorsement

ff137:
before publishing it to the trust registry.

Should allow regression mode to pass (previously it tried to fetch schema which was on trust registry, but not on ledger)


cl0ete (approved):
:rocket:

PR #1390, opened 2025-03-17
🚧 Add `schema_type` to trust registry

ff137:
🚧 Paused -- requires DB migration, so holding back on this one for the time being
___
Indy issuers can only use Indy schemas, and AnonCreds issuers can only use AnonCreds schemas.

So, our trust-registry should indicate which schema is of which type.


ff137:
Closing as no longer planned

PR #1393, opened 2025-03-18
:test_tube: Configure wallet-type in regression fixtures

ff137:
Correctly distinguishes Indy Issuer from AnonCreds Issuer in regression fixtures.


cl0ete (approved):
:rocket:

PR #1397, opened 2025-03-19
:test_tube: Test AnonCreds credentials e2e

ff137:
Adds connection fixtures for AnonCreds issuers, and adds tests for:
- connecting using public did
- getting paginated credential exchange records
- different credential flows

Changes `issuer_id` in our AnonCreds credential model to `issuer_did`.
Now optional: defaults to using issuer's public did if not specified.


cl0ete (approved):
:rocket:

PR #1399, opened 2025-03-20
:test_tube: Add AnonCreds credential fixtures

ff137:
Deduplicates existing indy credential fixtures, to add fixtures for issuing one or many anoncreds, revoking many anoncreds, etc

:boom: Yet again breaks existing regression tests by renaming long-lived credential, in order to distinguish indy from anoncreds


cl0ete (approved):
GGWP :rocket:

PR #1401, opened 2025-03-20
:test_tube: Test AnonCreds revocation

ff137:
Deduplicates existing tests to run for both indy and anoncreds scenarios

There are some difference between available functionality for either case, so these tests will help elucidate that


cl0ete (changes requested):
Clear pending not implemented for anoncreds. The rest looks like it should work :rocket:

cl0ete (approved):
LGTM

PR #1423, opened 2025-03-26
:construction_worker: Append SHA to manual builds

rblaine95:
* When manually triggering a build, only `:branch` is tagged
* Tweak the docker metadata so it does `:branch-sha`


ff137 (approved):
👏 Gremlins evicted

PR #1432, opened 2025-03-28
✅  Anoncreds test for Self-attested/predicate proofs 

cl0ete:
Extended `AnoncredsPresentationRequest` to give default values to `name` and `version` in model. Fix for test that was skipped


ff137 (approved):
Duplication warnings are fine for tests, and the 1 failing test is for regression that needs to be reset 👍

PR #1463, opened 2025-04-16
📝 e2e test docs

cl0ete:
Add break down of all e2e test.

also indicating some missing test


ff137 (commented):
Looks great!

One thing we can try do separately, is move e2e tests to their own folders, and then each respective doc can go with it.

e.g. we have folders for issuer and verifier tests, but not other cases - so we can make a folder for connections, and move all connection-specific e2e tests to that folder
That can be done separately

PR #1479, opened 2025-04-24
🐛 Fix connect processor

cl0ete:
`created_at` and `updated_at` is null on local (not sure about the rest)

- Use `this` to get events  `created_at` and `updated_at` time.

- Use `ts_unix_micro` (Looks like time stamps have 6 digits after decimal point )


ff137 (commented):
Makes sense 👍 I'll revert to @henrymsiska for approval

henrymsiska (approved):

PR #1502, opened 2025-04-30
:wrench: Set governance wallet-type to askar-anoncreds

ff137:
Just the config change. Separately will refactor creating of schemas to again only be allowed by governance, since that requires some test fixtures to change too


rblaine95 (approved):

rblaine95:
Does this require resetting any existing deployments?

wdbasson (approved):

PR #1515, opened 2025-05-05
:construction_worker: :wrench: Sync dependabot schedules

rblaine95:
* Saturday 07:00 for all dependabot actions


ff137 (approved):
I like. Better to end a week with dependabot PRs, than start with em

PR #1521, opened 2025-05-06
:rewind: Revert Python 3.13 upgrade

rblaine95:
There seems to be an issue with Uvicorn/FastAPI under load getting a SIGSEGV (segmentation fault) without any logs.

It's very inconsistent and happens after a while of throwing load at it (via k6)

@wdbasson

This is to test if reverting the Python 3.13 upgrade resolves the issue.

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -25,7 +25,7 @@ jobs:
           [tools]
           "pipx:black" = "25.1"
           "pipx:isort" = "6.0"
-          python = "3.13"
+          python = "3.12"
           EOF
 
       - name: Set up Mise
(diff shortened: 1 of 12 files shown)


rblaine95:
@wdbasson is gonna run some additional tests...
Probably also run overnight, just to be absolutely certain 🤞

ff137 (commented):
Hmm, would suck to have to downgrade again. So yeah, let's first confirm whether this is really necessary

cl0ete (approved):

ff137 (approved):
Downgrade due to sporadic SIGSEGV. Will debug separately

PR #1522, opened 2025-05-06
:boom: Remove all deprecations

ff137:
In preparation for the next major release, we should remove any outstanding deprecations.

💥:
- Remove deprecated options from DIDCreate
- Remove deprecated get_wallet_auth_token endpoint
- Remove deprecated credential_id from Exchange record
- Remove deprecated referent and record_id from wallet models 
  - (to be confirmed that the implementation here works as intended)

🎨:
- Un-deprecate credential /store endpoint
  - was marked deprecated by mistake
  - is still needed for OOB, and if something goes wrong with standard auto-store flow


cl0ete (approved):
Looks good! I am busy renaming the models and removing all uses of referent from our tests

PR #1561, opened 2025-05-14
:recycle: :test_tube: Refactor connection fixtures

ff137:
Replaces test fixtures that connect to issuers using their public DID to instead connect using their trust registry OOB invitation

Serves as a stepping stone to supporting did:cheqd, since it's currently not possible to connect using a cheqd public DID


ff137:
Regression test failures are expected, as we've modified how existing connections are fetched/established

PR #1574, opened 2025-05-20
✅ Update fixtures and tests

cl0ete:
`cred_defs` and `schema` now made with api call to `tenant-web`

### DIFF
--- a/app/tests/fixtures/definitions.py
+++ b/app/tests/fixtures/definitions.py
@@ -85,7 +85,10 @@ async def get_clean_or_regression_test_schema(
                 name, faber_client, governance_client
             )
         )
-    return schema_definition_result  # pylint: disable=possibly-used-before-assignment
+    else:
+        raise ValueError(f"Bad test mode: {test_mode}")
+
+    return schema_definition_result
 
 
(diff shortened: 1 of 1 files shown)


app/tests/fixtures/definitions.py
```diff
@@ -157,15 +173,18 @@ async def get_clean_or_regression_test_cred_def(
...
+        result = response.json()
     elif test_mode == TestMode.regression_run:
         result = await fetch_or_create_regression_test_cred_def(
-            auth=auth, schema=schema, support_revocation=support_revocation
+            client=client, schema=schema, support_revocation=support_revocation
         )
-    return result  # pylint: disable=possibly-used-before-assignment
+    return CredentialDefinition.model_validate(
+        result  # pylint: disable=possibly-used-before-assignment
+    )
```
ff137, line 187: We can remove the `# pylint: disable=possibly-used-before-assignment` / Unbound variable by adding after the elif:
```py
else:
    raise ValueError(f"Bad test mode: {test_mode}")
```

ff137 (approved):

PR #1584, opened 2025-05-28
:construction: :sparkles: Tails on S3

rblaine95:
:construction: :construction: :construction: :construction: :construction:


rblaine95:
> So there's no modification necessary to the acapy base image, and we're just masking the tails endpoints?

Precisely

PR #1597, opened 2025-06-02
:bug: Add duplicate file validation to put_file_by_hash

rblaine95:
Before uploading a file, check if a file with the same hash already
exists in the S3 bucket. If it does, return a 409 Conflict status
code with an appropriate error message.

This restores the original Tails Server behavior of preventing
duplicate file uploads based on hash values. The validation occurs
before any file processing to avoid unnecessary work.

* Check file existence using S3 head_object operation
* Return 409 Conflict if file already exists
* Handle S3 client errors appropriately
* Maintain compatibility with original server behavior

---

Update tool versions in mise configuration and Helm charts while
removing unnecessary security context overrides that weaken the
default security posture.

* Update istioctl from 1.25 to 1.26
* Simplify helm version specification to major version 3
* Upgrade Datadog Python library from v3.6.0 to v3.8.1
* Remove `readOnlyRootFilesystem: false` from security contexts
  to allow secure defaults in governance-agent, mediator, and
  multitenant-agent configurations


ff137 (approved):
LGTM

dockerfiles/app/Dockerfile
```diff
@@ -5,16 +5,16 @@ WORKDIR /app
...
-ENV POETRY_CACHE_DIR=/tmp/.cache/pypoetry
+ENV POETRY_CACHE_DIR=/poetry
 RUN poetry config cache-dir "$POETRY_CACHE_DIR" && \
   poetry config virtualenvs.path "$POETRY_CACHE_DIR/virtualenvs" && \
   poetry install --only main,app
 
-COPY app /app
-COPY shared /shared
+COPY --chown=65534:65534 app /app
+COPY --chown=65534:65534 shared /shared
```
rblaine95, line 17: This is required for local dev tilt live reloading.

The running user in the image needs to be the owner of the files which tilt live reloading overrides.
Otherwise need to run everything as root for local dev.

Everywhere that isn't local dev we have `securityContext.readOnlyRootFilesystem=true`

PR #1619, opened 2025-06-09
:pushpin: Pin Poetry install by hash

ff137:
Resolves OpenSSF pinned-dependencies warning for these docker files


cl0ete (approved):

.dockerignore
```diff
@@ -5,9 +5,10 @@
 !pyproject.toml
 !poetry.lock
 !app/
-!shared/
+!dockerfiles/poetry-requirements.txt
```
rblaine95, line 8: Does the Dependabot config need to be updated to point at this `poetry-requirements.txt`?
If it does, do we want it to be?

PR #1645, opened 2025-06-12
✨ Revocation status in list creds

cl0ete:
Add revoked status to list creds


app/models/verifier.py
```diff
@@ -15,6 +16,14 @@
 from shared.exceptions import CloudApiValueError
 
 
+class Status(str, Enum):
+    VALID = "valid"
+    REVOKED = "revoked"
+    NON_REVOCABLE = "non-revocable"
+    NOT_CHECKED = "not-checked"
+    CHECK_FAILED = "check-failed"
```
ff137, line 24: I think let's rename it to "RevocationStatus", and then we remove "non-revocable" as an option. So that RevocationStatus = None implicitly means it's non-revocable.

app/models/verifier.py
```diff
@@ -15,6 +16,14 @@
 from shared.exceptions import CloudApiValueError
 
 
+class Status(str, Enum):
+    VALID = "valid"
```
ff137, line 20: We can choose between "valid" or "active"

app/tests/services/wallet/test_wallet_credentials.py
```diff
@@ -0,0 +1,354 @@
+from unittest.mock import Mock, patch
```
ff137, line 1: I think we should have an e2e test as well - 
perhaps just latch on to an existing revocation test (because revocation tests are slow). Add an assertion (perhaps before and after revocation) that this endpoint's response matches "valid" and then "revoked"

For the non-revocable check we can add a new e2e test, because issuing normal creds is fast.

app/models/verifier.py
cl0ete: I like active more

PR #1660, opened 2025-06-17
:sparkles: Add mypy type checking

ff137:
Fixes a lot of type hint errors (about ~250 errors fixed before I started) - most important ones are for handling when a value can be None. We should assert a value is not None before trying to call actions on it.

Tries to make "cred_rev_id" type consistent - since it's always an int, but acapy sometimes returns or expect it to be input as a string.

Adds mypy to mise tasks and to github format workflow
___
For documentation purposes, here are some reasons why some are manually ignored (with `# type: ignore`):
- "Depends" / dependency injection cloaks expected types (when Depends is used as a default type in method args)
- Assigning vals in an if-else block, where first assignment the val is expected, but 2nd assignment (in else block) the val is optional - mypy flags this as an error
- Unpacking kwargs to a function can raise mypy errors, even though we know the methods args are correct
- Bad type hints from other libraries (e.g. nats-py accepts "start_time" arg as a string, but it's not indicated)
- Some complex while loops, which would raise if a value is not returned, mypy can't detect that returning None isn't possible

^ These are all cases where I think we should just ignore mypy errors, but other than that, we should try fix type errors as much as possible

___
Side note: I recently learned about `# pragma: no cover`, which we can add to avoid coverage being missing for impossible scenarios


cl0ete (approved):

.mise.toml
```diff
@@ -132,8 +132,13 @@ ruff format --check
...
   ruff format --check --diff "$file"
 done
+mypy . --ignore-missing-imports --check-untyped-defs --install-types --non-interactive
 """
 
+[tasks."mypy"]
+description = "Run mypy"
+run = "mypy . --ignore-missing-imports --check-untyped-defs --install-types --non-interactive"
+
 [tasks."bootstrap:kubeconfig"]
```
rblaine95, line 142: I think this should be renamed to `[tasks.lint]`
Like we've got `[tasks.fmt]` and `[tasks."fmt:check"]`

.mise.toml
```diff
@@ -132,8 +132,13 @@ ruff format --check
 find . -type f -name "Tiltfile" | while read -r file; do
   ruff format --check --diff "$file"
 done
+mypy . --ignore-missing-imports --check-untyped-defs --install-types --non-interactive
 """
 
+[tasks."mypy"]
```
rblaine95, line 138: I think this should be renamed to `[tasks.lint]`
Like we've got `[tasks.fmt]` and `[tasks."fmt:check"]`

.mise.toml
```diff
@@ -132,8 +132,13 @@ ruff format --check
 find . -type f -name "Tiltfile" | while read -r file; do
   ruff format --check --diff "$file"
 done
+mypy . --ignore-missing-imports --check-untyped-defs --install-types --non-interactive
```
rblaine95, line 135: I don't think this should be in the `fmt:check` task
```suggestion
```

.mise.toml
```diff
@@ -127,13 +127,29 @@ done
...
+depends = ["lint:ruff", "lint:mypy"]
+
+[tasks."lint:ruff"]
+description = "Only run ruff static checks"
+run = "ruff check"
+
+[tasks."lint:mypy"]
+description = "Only run mypy type checks"
+run = """
+mypy . --ignore-missing-imports \
```
rblaine95, line 148:
```suggestion
run = """
#!/bin/bash
mypy . --ignore-missing-imports \
```

rblaine95 (approved):

PR #1666, opened 2025-06-20
:building_construction: Refactor P2P networking for multi-replica LB

rblaine95:
* Extract LoadBalancer service to dedicated template with dynamic
  port allocation for each StatefulSet replica
* Remove P2P port from main and headless services since each replica
  now uses unique ports (26650 + replica index)
* Add RBAC permissions for service account to read LoadBalancer
  service metadata when `lbService.enabled` is true
* Implement init container to discover LoadBalancer hostname and
  share with main container via emptyDir volume
* Update cheqd-noded startup to dynamically configure P2P listener
  and external address based on replica ID and LoadBalancer hostname
* Change base P2P port from 26656 to 26650 to accommodate port
  range allocation across replicas
* Replace static `lbService.ports` with flexible `lbService.extraPorts`
  configuration option

This enables proper LoadBalancer support for P2P networking in
multi-replica deployments by ensuring each replica gets a unique
external port and can advertise the correct external address to
the P2P network.


ff137 (approved):
It's Greek to me, but it looks good and I trust you 😄

wdbasson (approved):

PR #1667, opened 2025-06-20
:zap: Check Tails File header early

rblaine95:
Improve performance by validating the AnonCreds tails file header
('00 02') during the first chunk read instead of after processing
the entire file. This eliminates the need to seek back to the
beginning of the temporary file for validation.

Changes:
* Move header validation from post-processing to first chunk read
* Add `first_chunk` flag to track initial chunk processing
* Update docstrings to be more specific about AnonCreds tails files

This optimization reduces I/O operations and provides faster feedback
for invalid files, improving overall upload performance.

### DIFF
--- a/tails/routers/tails.py
+++ b/tails/routers/tails.py
@@ -34,7 +34,7 @@ def get_s3_client() -> BaseClient:
 async def get_file_by_hash(
     tails_hash: str,
 ) -> StreamingResponse:
-    """Stream file content from S3"""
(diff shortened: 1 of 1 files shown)


ff137:
PS: Regression test failures seem to all be "server disconnected" errors between mt-web and acapy. The other one e2e failure is for a webhook not arriving. So, doesn't seem related to these changes

PR #1682, opened 2025-06-26
:sparkles: Trust Registry: Use async SqlAlchemy sessions

ff137:
The SqlAlchemy `Session` instance has a note: 
```
The :class:`_orm.Session` is **not safe for use in concurrent threads.**.
```

This was picked up after noticing there are many pgpool error logs:
```
psql pid 109151: LOG: child exiting, 30 connections reached
```

So, this PR switches away from using synchronous DB sessions to async.

Some of the sync db session code is maintained for potential alembic migrations.

___

NB: `?sslmode=prefer` is not supported with asyncpg in the database URL (`POSTGRES_DATABASE_URL`).

It's a bit cumbersome. `POSTGRES_DATABASE_URL` with `?ssl_mode=prefer` is valid for the sync engine. But with asyncpg, the query param is not supported, and it wants a connect_args = {"ssl": True or False}. 
(I haven't found how to configure "preferred" in asyncpg.)
So I've added an env var for `POSTGRES_SSL_REQUIRED` (true or false) specifically to configure the async connection.

### DIFF
--- a/trustregistry/database.py
+++ b/trustregistry/database.py
@@ -13,7 +13,7 @@ POSTGRES_DATABASE_URL = os.getenv(
 # Async engine for application use - automatically add asyncpg driver
 POSTGRES_ASYNC_DATABASE_URL = POSTGRES_DATABASE_URL.replace(
     "postgresql://", "postgresql+asyncpg://", 1
-)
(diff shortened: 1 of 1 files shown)


cl0ete (approved):
Nice

PR #1696, opened 2025-07-01
:construction: Revocation Service

ff137:
Dedicated service for managing the complexities around creating revocation registries and publishing revocations.


ff137:
Paused. Maybe move the separate service to a singleton class in the app module

PR #1708, opened 2025-07-02
:bug: Waypoint: only ack messages that match filter

ff137:
The reason the k6 "create-invitation" scenario continues to fail sporadically, is because Waypoint ack's every NATS message that matches the topic/state being listened to. We should only ack a message if it matches the field/field_id filter as well -- this will allow all events to still be available for other field/field_id requests.

This PR modifies the event generator to return both the Event with the NATS Msg object, so the latter can be acked upstream if it matches the filter, instead of ack'ing every event in the generator itself.


cl0ete (approved):

ff137:
Appears that the k6 issue is to do with using `their_did`. To be resolved in #1710

PR #1710, opened 2025-07-03
✅ Test sse events and update k6

cl0ete:
Test getting events multiple times 

When creating connections in k6 wait for events on `invitation_msg_id` field

### DIFF
--- a/app/tests/e2e/test_sse.py
+++ b/app/tests/e2e/test_sse.py
@@ -49,10 +49,7 @@ async def test_sse(
     wallet_id = bob_tenant.wallet_id
     sse_tasks = []
     for i in range(100):
-        task = (
-            await get_event_data(bob_member_client, wallet_id, f"test_sse_{i}"),
-            (f"Failed to get event data on attempt {i}"),
-        )
+        task = get_event_data(bob_member_client, wallet_id, f"test_sse_{i}")
         sse_tasks.append(task)
 
     results = await asyncio.gather(*sse_tasks)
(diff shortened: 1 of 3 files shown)


app/tests/e2e/test_sse.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+@pytest.mark.anyio
+async def test_sse(
+    bob_member_client: RichAsyncClient,
+    bob_tenant: CreateTenantResponse,
+    alice_member_client: RichAsyncClient,
+):
+    connection_tasks = []
+    for i in range(100):
```
ff137, line 41: I would also just replace the magic number with a var, since it's used in line 51 too.

Can maybe set it to 20 as a final value - something more modest / reasonable

scripts/k6/scenarios/create-invitations.js
```diff
@@ -190,6 +191,8 @@ export default function (data) {
       },
     });
 
+    const { invitation_msg_id: invitationMsgIdTemp } = JSON.parse(getHolderPrivateDidResponse.body);
+    invitationMsgId = invitationMsgIdTemp;
     const { my_did: holderPrivateDidFull } = JSON.parse(getHolderPrivateDidResponse.body);
     holderDid = holderPrivateDidFull.split(':').slice(0, 3).join(':');
     holderFullDid = holderPrivateDidFull;
```
ff137, line 198: Can maybe clean up the did assignments, in case it's not being used anymore

And so we should also just update the DIDExchange flow in the else block (lines 200 -- 228) -- and use invitation_msg_id there as well

ff137 (approved):

PR #1765, opened 2025-08-02
:arrow_up: Upgrade dependencies

ff137:
fastapi from 0.115 to 0.116
boto3 from 1.39 to 1.40
sse-starlette from 2.3 to 3.0
mypy from 1.16 to 1.17


rblaine95:
I think you can cherry-pick https://github.com/didx-xyz/acapy-cloud/pull/1763 in here as well

PR #1813, opened 2025-08-26
✨ Registry events monitor

cl0ete:
Added connect-processor `registry-events`: monitors registry events and saves state to `NATS kv`

Added connect-processor `registry-monitor`: looks at `NATS kv` and send message to jetstream if issue


ff137 (approved):
:clap:

PR #1843, opened 2025-09-12
💥 Return existing connection to public DID

cl0ete:


ff137:
PS: These regression tests will need to be updated, since existing connection is now returned by default.
```
FAILED app/tests/e2e/test_did_exchange.py::test_create_did_exchange_request
FAILED app/tests/e2e/test_did_rotate.py::test_rotate_did
FAILED app/tests/e2e/test_did_rotate.py::test_hangup_did_rotation
```

We could just skip those tests in regression mode

